PR #23, opened 2021-06-10
Ipr dbl click

irosengr:
Fixed doupleclick=True argument usage. It was not working as expected with ClickText and ClickElement. Also added test cases.


tkoukkari (approved):
Ok

joonasinisalo (approved):

Issue #19, opened 2021-06-01
VerifyTextCount doesn't work if real count is zero and expected_count is more than zero


ansevala:
**Describe the bug**
VerifyTextCount always passes when text isn't found at all. It only works properly when text is found X times.

**To Reproduce**
1. Go to _https://en.wikipedia.org/wiki/New_Britain_friarbird_
2. VerifyTextCount of any text that is not on the page expecting it is there, e.g.
```VerifyTextCount    Punatulkku    1```
3. Keyword passes

**Expected behavior**
Keyword should be failing.

tkoukkari:
Fixed, will be included in version **1.0.5** which will be released this week.

PR #25, opened 2021-06-10
fixes issue #19

tkoukkari:
VerifyTextCount will raise an exception / fail if text is not found


joonasinisalo:
Do we have an acceptance test case that verifies this?

tkoukkari:
> Do we have an acceptance test case that verifies this?

Good point, added one

joonasinisalo (approved):

irosengr (approved):

PR #37, opened 2021-09-10
Tko mac edge support

tkoukkari:
+ adding Edge tests to Mac & Windows pipeline


tkoukkari:
Edge tests added to Windows and Mac pipeline. GitHub does not seem to have Edge in Linux image yet.

joonasinisalo (approved):
There could be if-elif structure in edge.py when variating between platforms instead of three ifs.

irosengr (approved):

PR #44, opened 2021-10-19
Tko october fixes

tkoukkari:
### Added
- Configuration option **IsModalXpath** to limit text based search to elements under specific (modal) element
- Keywords **IsItem** and **IsNoItem**
- Added excludeSwitches option to Edge & Chrome, should not log all sorts of unnecessary things to console
- Added --no-sandbox option when Edge is run in docker

### Fixed
- **LogPage** keyword
- Fixed: example in ClickText documentation does not display correctly

### Changed
- Security update to OpenCV dependency
- Imported functions that are not keywords removed from documentation
- Clarified ExcpectFileDownload documenation


irosengr (approved):

joonasinisalo (approved):

QWeb/internal/text.py
```diff
@@ -192,6 +202,26 @@ def _get_contains_text_element(text, **kwargs):
     return element.get_webelements_in_active_area(xpath, **kwargs)
 
 
+def _filter_by_modal_ancestor(elements):
+    xpath = CONFIG["IsModalXpath"]
+    if xpath.startswith("//"):
+        xpath = xpath[2:]
+
+    logger.debug(f"length before filterin: {len(elements)}")
```
Ekexyz, line 210: missing g on filtering

Ekexyz (approved):

QWeb/internal/text.py
tkoukkari: I will fix directly on release commit, since pipeline will run tests there anyway and this should not affect anything

PR #60, opened 2022-03-09
Updates to M1 Mac instructions

joonasinisalo:


tkoukkari (commented):
Just to confirm: this is ONLY needed on M1 Mac's which are not using the latest OS version (Monterey) yet. Right? Just installing from pip should work if you are in Monterey?

joonasinisalo:
> Just to confirm: this is ONLY needed on M1 Mac's which are not using the latest OS version (Monterey) yet. Right? Just installing from pip should work if you are in Monterey?

Needed for Monterey also!

tkoukkari (approved):

irosengr (approved):

PR #65, opened 2022-03-15
64 some keywords do not accept named first argument

turunenm:
Fix and acceptance tests

Should not break anything as the change attempts to resolve variable `locator` in method `_equal_sign_handler` and this variable is only used in `timeout_decorator` as an exception message detail.

Added resolution change and tag `RESOLUTION_DEPENDANT` to `swipe/Scroll To` because it was not passing in local environment with a 4k screen.
Fixed `element/Click Element by WebElement instance` which was using xpath instead of WebElement to target the action

### DIFF
--- a/QWeb/internal/decorators.py
+++ b/QWeb/internal/decorators.py
@@ -179,21 +179,22 @@ def _equal_sign_handler(args, kwargs, function_name):
         locator = args[0]
     except IndexError:
         for key, value in kwargs.items():
-            # if present these are always the first argument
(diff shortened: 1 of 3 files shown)


Ekexyz (approved):
Thanks @turunenm, looks good. Lets see if @tkoukkari has a merge day in mind :)

tkoukkari (approved):
Good improvement. This behavior has bugged me sometime in the past,

PR #69, opened 2022-03-23
Safari fixes

tkoukkari:
Fixes/changes these issues:
- Automatic Frame search for Safari
  - Own solution to traverse through frames, as switch_to.parent_frame() works differently in Safari than other browsers 
- Window Management for Safari
  - keeps track of open windows, since window_handles returns handles in random order 
- Support for mac COMMAND key
- Removed "not suitable for mac/safari" tags from most tests
- Added Safari to Mac pipeline

### DIFF
--- a/QWeb/internal/frame.py
+++ b/QWeb/internal/frame.py
@@ -28,7 +28,7 @@ from QWeb.keywords import config
 import QWeb.internal.frame_checker as fc
 from QWeb.internal.exceptions import QWebDriverError, QWebElementNotFoundError, \
                                      QWebTimeoutError, QWebBrowserError, FATAL_MESSAGES
-from QWeb.internal import xhr, browser
+from QWeb.internal import xhr, browser, util
 from QWeb.internal.config_defaults import CONFIG
 
 
@@ -154,6 +154,7 @@ def set_html_source_count(value):
     BuiltIn().set_global_variable("${html_source_count}", value)
 
 
+# pylint: disable=R0915
 def all_frames(fn):
     """Decorator that takes any func as a parameter.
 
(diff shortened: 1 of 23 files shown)


irosengr (approved):

Ekexyz (approved):

QWeb/keywords/input_.py
```diff
@@ -683,10 +685,20 @@ def press_key(locator, key, anchor="1", timeout='0', **kwargs):
...
     """
+    driver = browser.return_browser()
+    action = ActionChains(driver)
     try:
         input_element = input_.get_input_elements_from_all_documents(
             locator, anchor, timeout=timeout, index=1, **kwargs)
         key = input_handler.check_key(key)
-        input_element.send_keys(key)
+
+        if key[0] == '\ue03d' and util.is_safari():  # COMMAND key workaround on safari
```
turunenm: Is it possible to use other mod keys together with CMD?

For example
`{SHIFT + \ue03d + a}`

If so, then it would be better to check if CMD is in the keys instead of key[0]
`if util.is_safari() and '\ue03d' in key:`

QWeb/keywords/input_.py
```diff
@@ -683,10 +685,20 @@ def press_key(locator, key, anchor="1", timeout='0', **kwargs):
...
+            javascript.execute_javascript("arguments[0].focus();", input_element)
+            action.key_down(key[0]) \
+                  .send_keys(key[1]) \
+                  .key_up(key[0]).perform()
```
turunenm, line 702: There are at least two cases which will cause trouble

User sends the only the CMD key -> causes index error
User sends CMD + Other mod + key -> some key will not be sent

```
cmd = '\ue03d'
key.remove(cmd)
action.key_down(cmd) \
         .send_keys(key) \
         .key_up(cmd).perform()
```
It might be necessary to handle the case empty key list separately

QWeb/internal/frame.py
```diff
@@ -207,7 +208,63 @@ def search_from_frames(driver=None, current_frame=None):
...
+                web_element = fn(*args, **kwargs)
+            except QWebElementNotFoundError as e:
+                err = e
+                web_element = None
+            if is_valid(web_element):
+                return web_element
+            start = time.time()
+            timeout = CONFIG['FrameTimeout']
+            while time.time() < timeout + start:
+                frames = fc.check_frames(driver)
```
turunenm, line 239: This is duplicate code with `search_from_frames()`, consider refactoring

QWeb/internal/frame.py
```diff
@@ -207,7 +208,63 @@ def search_from_frames(driver=None, current_frame=None):
...
+                    except WebDriverException as e:
+                        driver.switch_to.default_content()
+                        raise e
+                    config.set_config('FrameTimeout', float(timeout + start - time.time()))
+                    logger.trace('Frame timeout: {}'.format(timeout))
+                if err:
+                    raise err
+                return web_element
+            driver.switch_to.default_content()
+            raise QWebTimeoutError('From frame decorator: Unable to locate element in given time')
```
turunenm, line 265: This is duplicate code `search_from_frames()`, consider refactoring

QWeb/internal/frame.py
```diff
@@ -207,7 +208,63 @@ def search_from_frames(driver=None, current_frame=None):
...
+            try:
+                web_element = fn(*args, **kwargs)
+            except QWebElementNotFoundError as e:
+                err = e
+                web_element = None
+            if is_valid(web_element):
+                return web_element
+            start = time.time()
+            timeout = CONFIG['FrameTimeout']
+            while time.time() < timeout + start:
```
turunenm, line 238: I don't understand the purpose of this while loop.
At the end of the loop: L261-263
Either an exception is raised or a web_element is returned which should mean that this while-statement acts like if-statement in reality

turunenm (commented):
Check comments and consider making changes

QWeb/internal/frame.py
tkoukkari: It is safari specific function with few differences. Was left there in purpose. We first had only one and did the same for all browsers, but since switch_to.parent_frame() works correctly on all other browsers we decided to keep the old, working one for other browsers and introduce a safari specific solution only for safari.

Since it's recursive and we need to keep track on traversed frames in safari (parent_tree), we think it's easier to remember 6 months from now when there's strictly safari specific funtion.

QWeb/keywords/input_.py
tkoukkari: mainly meant for copy, paste, select all etc. what you can do in a (safari) browser text field. This should support those use cases. If other use cases arise, let's then handle those.

QWeb/internal/frame.py
tkoukkari: Very old code, not much documentation. I wouldn't dare to touch it without exceptionally good reason and lot's of debugging and testing.

The basic idea is to catch and raise specific exceptions to timeout decorators. Certain exceptions make it to re-try finding elements from all frames etc. Timeout guarantees we will not get stuck, i.e jumps out of the loop once timeout expires.

There are also **no** version of keywords where it's ok that webelement is not found.

QWeb/keywords/input_.py
tkoukkari: cmd alone will not make any sense, but we can just send it as any other key. Empty key list should not be possible, since then the "key" argument is not given. I'll check for length as well in CMD cases, we will support only "normal" cases that make sense in text field context for now (CMD+C, CMD+A etc.).

tkoukkari: New commit should resolve this

QWeb/internal/frame.py
tkoukkari: same comment as for above. It's intentionally safari specific.

turunenm (approved):
Looks good to me

PR #70, opened 2022-03-29
Rfw50

tkoukkari:
Fixes issues with filtering secrets from logs in robot fw 5.0

### DIFF
--- a/QWeb/internal/secrets.py
+++ b/QWeb/internal/secrets.py
@@ -58,6 +58,17 @@ def _replace_keyword_args(keyword, args):
         setattr(keyword.data, 'args', args)
 
 
+def _hide_keyword_arg_values(keyword):
+    par_index, secret = filtered_keywords[keyword.kwname]
+    censored_args = list(keyword.args)
+    if secret == 'hint':
+        censored_args[par_index] = 'SECRET'
+    else:
+        censored_args[par_index] = "*" * 5
+
+    return censored_args
+
+
 def _filtered_start_keyword(keyword):
     """Modify Robot FW internal function "start_keyword".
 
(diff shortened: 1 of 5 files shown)


QWeb/internal/secrets.py
```diff
@@ -38,6 +38,26 @@
...
 
 
+def _replace_keyword_args(keyword, args):
+    """Replace keyword args with the provided args
+
+    This function was added to support also RFW 4.0/5.0 RFW changes.
+    """
+    try:
+        keyword.args = args
+        return
```
turunenm, line 48: I think it would be better to check the version of used rfw and act accordingly.
It should make the code more explicit and understandable

```
from packaging import version
from robot.version import get_version
...
version_info = version.parse(get_version())
if version_info.major < 4:
    keyword.args = args
    return
```

QWeb/internal/secrets.py
```diff
@@ -47,23 +67,24 @@ def _filtered_start_keyword(keyword):
...
-        keyword.args = list(keyword.args)
+        censored_args = list(keyword.args)
         if secret == 'hint':
-            keyword.args[par_index] = 'SECRET'
+            censored_args[par_index] = 'SECRET'
         else:
-            keyword.args[par_index] = "*" * 5
+            censored_args[par_index] = "*" * 5
+
+        _replace_keyword_args(keyword, tuple(censored_args))
```
turunenm: This block of code is duplicated in `_filtered_end_keyword()`

Maybe turn it into a function?
`_hide_keyword_arg_values(keyword)`

turunenm (commented):
Code is functionally ok.
Consider making changes based on comments.

I will approve after the conversations have been resolved.

QWeb/internal/secrets.py
tkoukkari: @samijt as the same code is used in our other libraries, we should probably decide together and change it everywhere if wanted.

tkoukkari: Agreed to leave this as it is. Current solution works with every supported version.

QWeb/internal/secrets.py
tkoukkari: ok, will update soon.

tkoukkari: resolved by new commit

irosengr (approved):

turunenm (approved):
lgtm

PR #72, opened 2022-04-05
Full page screenshot

### LINKED ISSUE #61: Log Screenshot - Add optional parameter to capture full window
**Is your feature request related to a problem? Please describe.**
The [Log Screenshot](https://qentinelqi.github.io/qweb/QWeb.html#Log%20Screenshot) keyword currently only takes a screenshot of the visible window, so elements beyond the current scroll position are cut off from the screenshot.

**Describe the solution you'd like**
The [Log Screenshot](https://qentinelqi.github.io/qweb/QWeb.html#Log%20Screenshot) keyword should have an optional parameter, perhaps called `full_page`, that accepts a boolean value. If `full_page=True`, then QWeb should take a full page screenshot so that nothing is cut off from view. `full_page` should be set to `False` by default.

**Describe alternatives you've considered**
The other popular library for screenshots, [ScreenCapLibrary](https://mihaiparvu.github.io/ScreenCapLibrary/ScreenCapLibrary.html#Take%20Screenshot), does not seem to support this idea of full page screenshot either.

**Additional context**
Could QWeb support this or is this request outside of QWeb's available functionality with the web driver?

tkoukkari:
Implements issue #61 

~~We decided to enable this only for browsers that support full page screenshots natively (Firefox at the moment). We could take multiple screenshots and scroll in between + stitch the images together at the end, but this will not work too well if page has dynamic elements that move with scrolling.~~


kalyncoose:
Totally understandable, will work nicely in QWeb once all major browsers support that natively. Thank you for looking into it and implementing what you could!

tkoukkari:
> Totally understandable, will work nicely in QWeb once all major browsers support that natively. Thank you for looking into it and implementing what you could!

I will add Chrome & Edge support too. Found another way which seems to be good enough.

tkoukkari:
Example screenshots with Chrome:

**Normal:**
screenshot-RFDEBUG_REPL-e01e221e-8e9c-4af9-b338-8313ff9669b2

**Full page:**
screenshot-RFDEBUG_REPL-8d2562a5-cb2d-4cb6-8fb3-cdc55e566964

QWeb/internal/screenshot.py
```diff
@@ -272,3 +283,56 @@ def get_url():
...
+            'returnByValue': True,
+            'expression': script
+        })
+        return response['result']['value']
+
+    metrics = evaluate("({"
+                       "width: Math.max(window.innerWidth, "
+                       "document.body.scrollWidth, "
+                       "document.documentElement.scrollWidth)|0,"
+                       "height: Math.max(innerHeight, document.body.scrollHeight, "
```
turunenm: Should `innerHeight` be `window.innerHeight`?

turunenm (commented):
Looks good to me.
Check the comment if it requires a change.

QWeb/internal/screenshot.py
tkoukkari: Thanks, yes I think it should. Will resolve in a new commit.

turunenm (approved):
lgtm

irosengr (approved):

PR #76, opened 2022-04-14
75 add type checking

### LINKED ISSUE #75: Add type checking
**Is your feature request related to a problem? Please describe.**
At the moment figuring out some keyword/method inputs/output types requires the deducing types from the source code.
This is tedious and error prone.

**Describe the solution you'd like**
Add typing to all method/keyword definitions and check the types statically with mypy on commit

turunenm:
Added typing to all keywords & methods
This required refactoring here and there, mostly due to variable reassigments with different types

Added requirements_test.txt dependencies
- mypy
- types-requests
- types-setuptools

Added mypy configs to setup.cfg
Added type checking to GitHub Actions
Added new duty `typing`

Disabled icon.robot acceptance-tests from macos

PR is massive, sorry about that 😬


QWeb/__main__.py
```diff
@@ -29,12 +29,25 @@ def cli():
         '%(prog)s --list Get (lists all keywords startin with "Get")\n'
         '%(prog)s --show TypeText (displays documentation for keyword "TypeText")',
         description='Note: This module is meant to be used as a robot famework library.'
-        'Command line interface only provides access to keyword documentation.')
```
tkoukkari, line 32: I wonder why the formatter changed these? Maybe different line length setting or something. For me looks uglier when divided to (unnecessary) multiple lines

turunenm: Probably due to line length.
I think I need to check yapf line len setting and rerun the formatter...

turunenm: Setting line len to 100 with yapf mostly fixed this issue

QWeb/internal/browser/__init__.py
```diff
@@ -14,82 +14,82 @@
...
+_current_browser: WebDriver = None  # type: ignore[assignment]
+_open_browsers: list[WebDriver] = []
+
+
+def get_current_browser() -> WebDriver:
     return _current_browser
 
 
-def set_current_browser(index):
+def set_current_browser(index: str) -> None:
```
tkoukkari: Can't this be int as well? Or was it so that it always comes from robot script as string?

tkoukkari (approved):
Approved. Nice that you had added own "duty" for checking types!

Let's release 2.0.1 shortly (this week?) and then merge this -> target release **2.0.2**.

QWeb/internal/browser/__init__.py
turunenm: True that it could also be int.
Problem is that the first check of the function, `if index.isdigit():`, causes an AttributeError if an int is passed as a parameter

This function was only called from `keywords/browser.py - switch_browser` where the parameter is cast to str

Refactored the functions to allow passing param `index` as an int
I changed the type to `Union[int, str]` and removed the str cast from the `switch_browser` kw.

Commit
[d18ce3f](https://github.com/qentinelqi/qweb/pull/76/commits/d18ce3f625e6568b5da423a830f0fa9bc6a0281d)

irosengr (approved):

PR #81, opened 2022-05-12
74 verifyelementtext doesnt highlight the elementtext under test when running the test case

### LINKED ISSUE #74: VerifyElementText doesn't highlight the elementtext under test when running the test case
When running test with keyword VerifyElementText it doesn't highlight the element text under test. It would be improvement to see and confirm that the keyword tests certain element text with anchor (because the same text can be multiple times in different places on the page).  

Example:
`VerifyElementText    ipsum    ipsum    5    dolore`

`    SetConfig    SearchMode           Draw`
Doesn't work either with VerifyElementText

turunenm:
Includes a fix and a new acceptance test in _element.robot_
Placed the test under _element.robot_ since this tests the `draw border feature` and not the `VerifyElementText`

### DIFF
--- a/test/acceptance/element.robot
+++ b/test/acceptance/element.robot
@@ -1,6 +1,8 @@
 *** Settings ***
 Documentation                   Tests from element keywords
 Library                         QWeb
+Library    DatabaseLibrary
+Library    OperatingSystem
 Suite Setup                     OpenBrowser                 file://${CURDIR}/../resources/text.html                 ${BROWSER}           --headless
 Suite Teardown                  CloseBrowser
 Test Timeout                    1min
@@ -135,8 +137,6 @@ GetAttributeOK
     # with index using css
     ${attribute}                GetAttribute                input   value        element_type=css    index=5
     should be equal             ${attribute}                Show hidden
-    
-
 
 GetAttributeNOK
     [Tags]                      GetAttribute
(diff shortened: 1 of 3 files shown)


test/acceptance/element.robot
```diff
@@ -247,4 +246,23 @@ VerifyAttributeNOK
...
     # multiple found css
-    Run Keyword And Expect Error      QWebValueError:*              VerifyAttribute      button               value                Button2              element_type=css     timeout=2
\ No newline at end of file
+    Run Keyword And Expect Error      QWebValueError:*              VerifyAttribute      button               value                Button2              element_type=css     timeout=2
+
+VerifyElementTextBorders
+    [Tags]                      VerifyElementText    DrawBorders
+    Go To                       file://${CURDIR}/../resources/text.html
+    
+    ${no_border_file}           SetVariable            ${CURDIR}/no_border.png
```
tkoukkari, line 255: I can't see these images, should they be included?

Also maybe could just use VerifyIcon here?

turunenm: Those are just filepath strings for screenshots which will be taken during the the Test and there should be no need for stored images

turunenm: To clarify, test takes screenshot from initial state with no element highlights.
Then an element is highlighted and a new screenshot is taken.
The screenshots are then compared on binary level and they should be different.

VerifyIcon would also work but requires an existing image.

I don't actually know which would be better approach 😅

tkoukkari: ah, okay. That explains why I couldn't find those images :).

tkoukkari (approved):
lgtm

Ekexyz (approved):

PR #82, opened 2022-05-17
Rework GitHub Actions

turunenm:
Split test workflows into different files, _static_analysis.yml_, _[os]\_acceptance.yml_ and _tests_reusable.yml_

_static_analysis.yml_ runs type checking and linting, these used to be part of _[os]\_acceptance.yml_ and were ran 18 times instead of once.
_[os]\_acceptance.yml_ now only contain matrix variable definitions which are passed to _tests_reusable.yml_
_tests_reusable.yml_ contains generic unit and acceptance test workflow which works on all OS platforms (though limited to X-latest OS:s)

### DIFF
--- a/.github/workflows/linux_acceptance.yml
+++ b/.github/workflows/linux_acceptance.yml
@@ -15,69 +15,11 @@ on:
 
 jobs:
   linux_acceptance:
-
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [3.7, 3.8, 3.9]
-        browser: [chrome, firefox]
-
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
-      with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install RFW 3.2.2
-      if: matrix.python-version == '3.7'
-      run: |
-        python -m pip install robotframework==3.2.2
-    - name: Install dependencies
-      run: |
-        sudo apt-get install matchbox scrot
-        python -m pip install --upgrade pip
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-        if [ -f requirements_test.txt ]; then pip install -r requirements_test.txt; fi
-    - name: Display Python version
-      run: |
-        python -c "import sys; print(sys.version)"
-    - name: Display Robot FW version
-      run: |
-        python -m pip show robotframework
-    - name: Typecheck with mypy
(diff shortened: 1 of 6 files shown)


tkoukkari:
Only question:
- If something fails, does it cancel ALL other jobs or only jobs from that same platform as previously?
- Same with re-runs as well

Those were intentionally in separate files, since there are A LOT more issues with Windows/Mac runners than Linux runners in Github.

tkoukkari:
Actually second question as well. Can we still get status badges by platform as previously:
image

turunenm:
> Only question:
> 
>     * If something fails, does it cancel ALL other jobs or only jobs from that same platform as previously?
> 
>     * Same with re-runs as well
> 
> 
> Those were intentionally in separate files, since there are A LOT more issues with Windows/Mac runners than Linux runners in Github.

Failed execution is not reflected to other executors (even within the platform).
This can also be changed with `fail_fast: false`
```
strategy:
      fail-fast: false
      matrix:
        ...
```

With re-runs you can choose to _Re-run all jobs_ or _Re-run failed jobs_
If you want to manually execute all tests of a specific platform you can use `workflow_dispatch`



> Actually second question as well. Can we still get status badges by platform as previously: image

In this PR all of the os/python/browser combinations are handled as jobs instead of separate workflows.
Currently badges can not be assigned for jobs so the os-specific badges are not possible with this style.

turunenm:
Removed the chained execution of _Static Analysis -> OS-tests_ 
Now the _Static Analysis_ and the _OS-tests_ all start at the same time and failed _Static Analysis_ does not block _OS-tests_

Changed _os_acceptance.yml_ files to use the _tests_reusable.yml_ to lessen code duplication

Now the workflows behave like previously with one difference:
_Static Analysis_ is executed in it's own workflow and only once. Previously static analysis was part of _os_acceptance.yml_ workflows and was executed total of 18 times

tkoukkari:
> 

Ok, I would rather keep OS specific badges.

turunenm:
> > 
> 
> Ok, I would rather keep OS specific badges.

Badges work like they used to (since my previous comment)

Should be good for review 👍

tkoukkari:
> > > 
> > 
> > 
> > Ok, I would rather keep OS specific badges.
> 
> Badges work like they used to (since my previous comment)
> 
> Should be good for review 👍

Ok, great. Then no objections :)

tkoukkari (approved):

Ekexyz (approved):

irosengr (approved):

PR #83, opened 2022-05-18
Fixing get_item_using_anchor

mehent:
Issue: Click Item keyword fails with error message "UnboundLocalError: local variable 'correct' referenced before assignment" if SearchMode is set to None with RF built-in variable ${NONE}.
Fix: Assigning value to var 'correct' before ifCONFIG['SearchMode'] block.


tkoukkari:
@mehent Thank you! I think we've had a default value in SearchMode for so long, that this just hasn't come up.

turunenm (approved):
Looks good to me

Ekexyz (approved):

PR #84, opened 2022-05-23
Browserstack capabilities change

tkoukkari:
- change from legacy format to W3C.
-  Also now we support rest of browserstack capabilities as kwargs
- Updated browser versions to 'latest'
- log session id in openbrowser

### DIFF
--- a/QWeb/internal/browser/bs_desktop.py
+++ b/QWeb/internal/browser/bs_desktop.py
@@ -3,6 +3,7 @@ import os
 from selenium.webdriver.remote.webdriver import WebDriver
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException
+from typing import Any
 from robot.api import logger
 from robot.libraries.BuiltIn import BuiltIn
 from QWeb.internal import browser, exceptions
@@ -25,12 +26,12 @@ OS: dict[str, str] = {
 }
 
 
-def open_browser(bs_browser: str, project_name: str, run_id: str, **kwargs: str) -> WebDriver:
+def open_browser(bs_browser: str, project_name: str, run_id: str, **kwargs: Any) -> WebDriver:
     bs_key = BuiltIn().get_variable_value('${APIKEY}') or os.environ.get('bskey')
     bs_user = BuiltIn().get_variable_value('${USERNAME}') or os.environ.get('bsuser')
     bs_os = BuiltIn().get_variable_value('${BSOS}') or 'windows'
 
-    desired_caps = {
+    desired_caps:dict = {
         'bstack:options': {
             "buildName": project_name,
             "projectName": project_name,
--- a/QWeb/internal/browser/bs_desktop.py
+++ b/QWeb/internal/browser/bs_desktop.py
@@ -10,41 +10,49 @@ from QWeb.internal import browser, exceptions
 NAMES: dict[str, tuple[str, str]] = {
     # Default  versions for different browsers.
     'chrome': ('Chrome', 'latest'),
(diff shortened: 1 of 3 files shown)


QWeb/internal/browser/bs_desktop.py
```diff
@@ -3,49 +3,59 @@
...
 OS: dict[str, str] = {
     # Default versions for Windows and OSX.
-    'osx': 'Catalina',
+    'osx': '"Big Sur',
     'windows': '10'
 }
 
 
-def open_browser(bs_browser: str, project_name: str, run_id: str) -> WebDriver:
+def open_browser(bs_browser: str, project_name: str, run_id: str, **kwargs: Any) -> WebDriver:
```
turunenm, line 29: Most of the kwargs in the codebase do not have the `Any` type explicitly specified as it is the default type for kwargs.

This is matter of preference, but I think it would be good to decide do we want to add explicit typing for `kwargs: Any`
For the sake of uniformity it is a good idea but I was feeling slightly lazy while doing the initial typing and decided not to add the explicit typehints to `kwargs: Any`

I think I just answered to my own question and will be adding typing to all `kwargs: Any` -parameters 😅

turunenm, line 29: This is ok as is but there is a scenario to be aware of:
Python allows kwarg values to be objects and this typehint specifies that it is ok to pass `Any` type of value 
AFAIK browserstack options is `Dict[str:str]` meaning both the keys and values must be strings. 

This might blow up if user of `Open Browser` feels adventurous and tries to pass dictionary which contains non-string value like `{'a_key': [0,1,2]}`

If this typehint is set to `**kwargs: str` then mypy will complain at _keywords/browser.py - open_browser()_ because there it is reasonable that the kwargs contain objects like dictionaries with `prefs` kwargs.

Making typing here more strict is not worth the effort, but adding a comment to the source code about the actual typing of the kwargs would be nice

QWeb/internal/browser/bs_mobile.py
```diff
@@ -2,28 +2,40 @@
...
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException
+from typing import Any
 from robot.api import logger
 from robot.libraries.BuiltIn import BuiltIn
 from QWeb.internal import browser, exceptions
 
 
-def open_browser(bs_device: str, project_name: str, run_id: str) -> WebDriver:
+def open_browser(bs_device: str, project_name: str, run_id: str, **kwargs: Any) -> WebDriver:
```
turunenm, line 11: See _bs_desktop.py - open_browser()_ comments

turunenm (approved):
lgtm

QWeb/internal/browser/bs_desktop.py
tkoukkari: Actually I checked how it was done in some other part of the code and followed that. I don't have strong preference here.

tkoukkari: We can change this to 
`Dict[str:str]`
...if you want.

irosengr (approved):

QWeb/internal/browser/bs_desktop.py
tkoukkari: ah, re-read this and that was not actually what you were suggesting.

turunenm: Let's leave these kwarg typehints in place and I'll go through the typing of args and kwargs in another PR

Ekexyz (approved):

PR #94, opened 2022-06-21
changed screenshot filename handling

tkoukkari:
- generated filenames are limited to take max 100 chars from test name
- User given filenames are not limited, but help text was updated


irosengr (approved):

irosengr (approved):

QWeb/internal/screenshot.py
```diff
@@ -65,7 +66,8 @@ def _remove_invalid_chars(text_to_check: str) -> str:
     :param text_to_check:
     :return:
     """
-    return "".join(c for c in text_to_check if c in VALID_FILENAME_CHARS)
+    valid = "".join(c for c in text_to_check if c in VALID_FILENAME_CHARS)
+    return valid if len(valid) <= MAX_LENGTH else valid[:MAX_LENGTH - 1]
```
turunenm, line 70: Truncating the filename here can cause confusion in future because that is not part of what the function name describes.

Maybe create a separate function which is named something like `_truncate_too_long_filename(filename: str) -> str` and call it explicitly where needed.

turunenm (commented):
Solution is good but consider moving the filename truncation to a separate function.
I will approve the PR after the conversation is resolved

QWeb/internal/screenshot.py
tkoukkari: I disagree. There should be only one place where valid filename is returned from. Too long is considered invalid too.

We may think of renaming the function in the future.

turunenm: Ok, I'll resolve the conersation and approve the PR

turunenm (approved):
lgtm

PR #97, opened 2022-08-02
added shadow dom support for GetText with attr&tag

tkoukkari:
Old one from before summer holidays. GetText will support "items" using tags with this one.


samijt (approved):
Changes ok

joonasinisalo (approved):

PR #100, opened 2022-08-30
Tko element kws index

tkoukkari:
Added support for using index argument with xpaths

### DIFF
--- a/QWeb/internal/util.py
+++ b/QWeb/internal/util.py
@@ -311,3 +311,11 @@ def escape_xpath_quotes(text):
     if '"' in text:
         return f"'{text}'"
     return f'"{text}"'
+
+
+def anchor_to_index(anchor):
+    try:
+        index = int(anchor) - 1
+    except ValueError:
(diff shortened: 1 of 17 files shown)


QWeb/internal/util.py
```diff
@@ -310,3 +311,11 @@ def escape_xpath_quotes(text):
     if '"' in text:
         return f"'{text}'"
     return f'"{text}"'
+
+
+def anchor_to_index(anchor):
```
turunenm: typing should be added to function definition

turunenm (changes requested):
Add typing to commented function and this one's good to go

QWeb/internal/util.py
tkoukkari: Oops, yep, will do.

tkoukkari: done with the latest commit

turunenm (approved):
lgtm

irosengr (approved):

PR #106, opened 2022-11-18
Suppress webdriver warnings

turunenm:
`StaleElementReferenceError` and `WebDriverException` messages from decorator `frame.all_frames()` are being logged as warnings even though these do not lead to test failures. These warnings probably raise unnecessary concerns for the end users and should be suppressed by default.

warning

Changing log level of these messages to debug instead of warn so that unnecessary stacktrace spam is not shown in the execution logs unless log level is set debug.


tkoukkari (approved):
good catch

PR #110, opened 2022-11-24
Initial

turunenm:
Retry loops should not spam into console or info level log.

Changed a few `logger.info()` calls to `logger.debug()` calls to remove unnecessary messages from console and info level

Change at line 135 should be paid attention to. That exception handling block raises after logging so it might be reasonable to log at info level because the loop should stop after the raise.

image

**This PR also includes a GitHub Action fix for ubuntu-latest currently missing Firefox & Geckodriver**


tkoukkari (approved):
Approved once the checks are finalized

PR #111, opened 2022-12-07
Fix unbound local error with text anchors

turunenm:
If config `MultipleAnchors==True` and no anchors are found in method `QWeb/internal/text.py - get_element_using_anchor()` then an `UnboundLocalError` raises due to missing handling for the condition.

Fixed by adding a check which now raises more appropriate exception `QWebElementNotFound` if no anchors are found.

Also includes a minor linting fix to `QWeb/internal/decorators.py - perform()`


tkoukkari (approved):
lgtm

PR #112, opened 2022-12-09
Fix to ShadowDOM dropdowns

turunenm:
Fixes handling of Salesforce CPQ ShadowDOM dropdowns
Unfortunately I was not able to reproduce the issue with local html files and the dropdown in case is always behind a login


tkoukkari (approved):
lgtm

PR #119, opened 2023-03-06
make sure normal elements are included in results

tkoukkari:
...when shadow dom is on, but remove staling ones

Marked this as "Draft" for now in case anyone has better ideas. Here we remove all staled elements form original "normal" search results and append elements found via shadow dom result, if they are different than what we already found during normal search.


Ekexyz (approved):

irosengr (approved):

turunenm (approved):
lgtm

PR #124, opened 2023-07-31
break from loop when matching attribute is found

tkoukkari:
Fixes issue that came up with Chrome 115 -> driver throwing an exception if same element is added to a list multiple times

### DIFF
--- a/QWeb/internal/javascript.py
+++ b/QWeb/internal/javascript.py
@@ -163,9 +163,11 @@ def get_by_attributes(elements: list[WebElement], locator: str,
                for (var j = 0; j < attrs.length; j++) {
                     if (attrs[j].value.trim() == locator) {
                         full.push(elems[i]);
+                        break;
                     }
                     else if (partial && attrs[j].value.includes(locator)){
                         part.push(elems[i]);
+                        break;
                     }
                }
             }


QWeb/internal/javascript.py
```diff
@@ -163,9 +163,11 @@ def get_by_attributes(elements: list[WebElement], locator: str,
                for (var j = 0; j < attrs.length; j++) {
                     if (attrs[j].value.trim() == locator) {
```
turunenm, line 164: Off-topic from PR but if we `value.trim()` should we also `locator.trim()` or not `trim()` either of them?

turunenm (approved):
lgtm,
check the comment for an off-topic insight

QWeb/internal/javascript.py
tkoukkari: I think the original idea has been that since we try to find full text, we trim extra spaces from html (there have been multiple cases where this has been a problem). 

Could be worth trimming locator as well, but it's not that easy or common to have leading & trailing spaces in rf.

samijt (approved):

QWeb/internal/javascript.py
turunenm: One scenario comes to mind where matching trimming would be necessary:

- get attribute (with leading/trailing ws) from element 
- do something
- verify no item with the retrieved attribute exists to confirm action

If the trim does not match for attr-value and attr-locator then this kind of sequence will misbehave, verify no item would always pass even if the element would be present.

In order to make it work one would need to know that only the attr-value is trimmed and that the attr-locator should be trimmed by user before passing it to the function/kw.

This shouldn't be that common and might not be worth changing though.

PR #125, opened 2023-08-01
fix for situation where input is found via label

tkoukkari:
...for under shadow dom


tkoukkari:
WIP: One other fix will be coming if we get access to customer env

samijt (approved):

PR #126, opened 2023-08-04
Tko aug dependency updates

tkoukkari:
- Updating to newer version of dependencies
- Pipeline forced to use working older EdgeDriver version on Linux
- Bug fixes to VerifyFile and Scroll
- Fix: support running QWeb directly from Python, for example:

```
from QWeb import QWeb
qweb = QWeb()
qweb.open_browser("https://www.w3schools.com/html/html_forms.asp", "chrome")

for i in range(100):
     qweb.type_text("First", f"John_{i}")
     qweb.refresh_page()
```

### DIFF
--- a/.github/workflows/tests_reusable.yml
+++ b/.github/workflows/tests_reusable.yml
@@ -71,6 +71,15 @@ jobs:
         xauth generate $DISPLAY .
         popd
 
+    - name: Install working edgedriver
+      if: runner.os == 'Linux'
+      run: |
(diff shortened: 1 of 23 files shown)


JoniSaarinen:
Just out of curiosity, why selenium is updated only to 4.9? And not to the latest 4.11?

tkoukkari:
> Just out of curiosity, why selenium is updated only to 4.9? And not to the latest 4.11?

Yes we will move to latest selenium a bit later. There are some code changes that need to be done before that.

irosengr (approved):

samijt (approved):

PR #136, opened 2023-12-20
tests: updating workflow actions to later versions

tkoukkari:
Just updating GitHub action workflows to later versions


tkoukkari:
will be added as part of bigger PR

Issue #139, opened 2024-02-07
QWeb tries to inject JQuery to web pages


mbchy:
**Description**
QWeb tries to inject JQuery to web pages. If website has a strict Content Security Policy setting, where injecting 3rd party scripts is not allowed, QWeb fills robot logs with error messages: 

> Unable to check AJAX requests due error: Message: javascript error: Failed to set the 'src' property on 'HTMLScriptElement': This document requires 'TrustedScriptURL' assignment.

The problem seems to lie in setup_xhr_monitor() function defined in QWeb/internal/xhr.py and called from keywords Open Browser and Open Window.

I would propose two possible solutions:

1. make it possible to switch off the xhr monitor functionality
2. log the error messages with a different loglevel

This would be helpful if it is not possible to change the CSP settings of the System Under Test.

**To Reproduce**
1. Using QWeb, Open Browser (chrome) to a web page with strict CSP (require-trusted-types-for 'script'; default-src 'self')
2. Use any keywords (ClickText, VerifyText, TypeText, ...)
3. End test, open log
4. Info level error messages should be printed in keyword logs

**Expected behavior**
1. Error messages should be logged with debug or trace level
2. User should be able to configure the xhr_monitor functionality (turn it off)

**Screenshots**

![kuva](https://github.com/qentinelqi/qweb/assets/8687364/a6799cab-4933-4724-9601-9eec15496a4c)

**System info:**
 - OS: Windows
 - Browser: chrome, edge
 - Version: 121.0   (bug should reproduce with versions >83)

- Python 3.10.9
- QWeb 2.0.7
- selenium 4.6.0

**Additional context**

About CSP & TrustedTypes:

https://web.dev/articles/trusted-types
https://developer.mozilla.org/en-US/docs/Web/API/TrustedScriptURL
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for

tkoukkari:
Thank you. Yes, those messages should be clearly logged with at minimum "debug" log level. We will change this. For the second part, wait function can be overridden with "[Set Wait Function](https://qentinelqi.github.io/qweb/QWeb.html#Set%20Wait%20Function)"

mbchy:
Thank you for quick reply @tkoukkari ! I will try monkey patching the wait function. I'm a bit confused though as `Open Browser` and `Open Window` seem to directly call `setup_xhr_monitor()` which tries to inject the script.

[...]

Issue #138, opened 2024-01-26
Let user decide if cookies are logged or not


tkoukkari:
This is off topic from the PR but I've been wondering what is the rationale to info logging the cookies automatically.
I have some scripts where I'm looking for cookie values to update on a rapid pace and it clogs up the report.

Maybe add argument `log_cookies=True` to provide control over it for the user?

_Originally posted by @turunenm in https://github.com/qentinelqi/qweb/pull/137#discussion_r1467249531_

tkoukkari:
Fixed on release 3.2.0

Issue #90, opened 2022-06-02
Clean up unit tests and acceptance tests


turunenm:
Tests which have been jailed for a long time should be completely removed.

Might be a good idea to clean up errors caused by deprecated kws etc.

PR #140, opened 2024-03-05
Tko feb 2024

tkoukkari:
### Fixes:

- #139 
- #138 
- #90 (to some extent, i.e. some tests were left as "jailed" for now but most were either removed or taken into full suite)
- **VerifyLinks** default timeout changed, wrong format was used previously
- Table keywords gave unhelpful error if **UseTable** was not used
- Many doc fixes

### New features:
- Adds "enforced mode" to SetConfig **SearchDirection**
- Adds "delay" kwarg support to **WriteText**
- Added "partial_match" support to **VerifyTable** and **Click Cell**


irosengr (approved):

QWeb/internal/element.py
```diff
@@ -360,7 +367,7 @@ def _calculate_closest_distance(element1: WebElement, element2: WebElement) -> f
     search_direction = CONFIG["SearchDirection"]
```
turunenm, line 367: i think it would be better to strip the possible exclamation mark and leave the if statements as they were.
In order to be sure what the search_directions can actually be one will need to find the validation method in search_strategy.py

```
search_direction = CONFIG["SearchDirection"]
# check and strip enforce char
if search_direction.endswith("!")
    search_direction = search_direction[:-1]
```

samijt (approved):

QWeb/internal/table.py
```diff
@@ -193,7 +198,8 @@ def get_clickable_cell(self,
             return clickable_child[int(index) - 1]
         return table_cell
 
-    def get_cell_by_locator(self, locator: str) -> int:
+    def get_cell_by_locator(self, locator: str, **kwargs) -> int:
+        partial_match = util.par2bool(kwargs.get('partial_match', CONFIG['PartialMatch']))
```
turunenm, line 202: Shouldn't this also check if `partial_match` was provided in the kwargs?

QWeb/keywords/config.py
```diff
@@ -112,8 +112,8 @@ def set_config(par: str, val: Any) -> Any:
     | ScreenShotType_     | Log html source, screenshot or both     | screenshot     |
     +---------------------+-----------------------------------------+----------------+
     | SearchDirection_    | Set relative search direction for       | closest        |
-    |                     | element search (up, down, lef, righ,    |                |
-    |                     | closest)                                |                |
+    |                     | element search (closest, up, down, left,|                |
+    |                     | right, !up, !down, !left, !right)       |                |
```
turunenm: Exclamation marks should be after the direction

QWeb/keywords/table.py
```diff
@@ -132,10 +144,13 @@ def verify_table(coordinates: str,
...
     """
-    table = Table.ACTIVE_TABLE.update_table()
-    if isinstance(ACTIVE_TABLE, Table) is False:
+    if not isinstance(ACTIVE_TABLE, Table):
         raise QWebInstanceDoesNotExistError('Table has not been defined with UseTable keyword')
-    table_cell = table.ACTIVE_TABLE.get_table_cell(coordinates, anchor)
+    table = Table.ACTIVE_TABLE.update_table()
+
+    table_cell = table.ACTIVE_TABLE.get_table_cell(coordinates, anchor, **kwargs)
+    partial_match = util.par2bool(kwargs.get('partial_match', CONFIG['PartialMatch']))
```
turunenm, line 152: Also check partial_match from kwargs?

turunenm (changes requested):
1 doc string change
1 suggestion
2 questions about the table partial_match kwarg

QWeb/internal/table.py
tkoukkari: wdym? It get's partial_match from kwargs if it's given, if not it takes it from CONFIG

QWeb/keywords/config.py
tkoukkari: damn it :)

QWeb/keywords/table.py
tkoukkari: as in previous one, it checks kwargs and if not given takes the value from CONFIG

QWeb/internal/table.py
tkoukkari: It is checking kwargs...get uses default value (i.e. value from CONFIG) if kwarg partial match is not given

QWeb/keywords/config.py
tkoukkari: oops, leftovers when we changed the format. Will update with a new commit

QWeb/keywords/table.py
tkoukkari: same thing here, it is checked. kwargs.get checks if "partial_match" is there, if not it defaults to CONFIG value of "PartialMatch"

QWeb/internal/table.py
turunenm: I don't know what I've been thinking 😅 
My bad, resolving comment.

QWeb/internal/element.py
tkoukkari: Not sure I get what you mean. the idea here is to actually fail if text is found but not from the direction the anchor (in enforced mode). Previously without enforced mode it did not fail, but selected the "closest". 

Thus we need different handling for enforced and normal mode. Also I want it to be visible in any logging that enforced mode was used -> not wise to strip the "!"

turunenm: The thing I was wondering when looking at the changes was "can values like 'something left and else' get passed in here" as they would pass the checks `if 'left' in search_direction:`

I thought about the stripping after the review and I'd actually prefer making the checks more explicit (instead of stripping) in this fashion:
`if search_direction in ('left', 'left!'):`

I'd also like to see a raising `else:` block with comment that the execution should never get there just to be sure and to remove doubts of the behavior in case if the code reader does not happen to look at the config input validation function.

tkoukkari: ok, that makes sense.

tkoukkari: the else branch is not really needed because value is checked when set and it can't be any other than one of those allowed values.

```
SetConfig                   SearchDirection     something left
! keyword: SetConfig                   SearchDirection     something left
! handler execution failed: ValueError: Wrong search direction
```

turunenm (approved):
lgtm

Issue #141, opened 2024-03-21
Can not retrieve text from some <slot> elements with Get Text or Get Cell Text


tkoukkari:
**Describe the bug**
If a slot element has just text in it (no child elements) it's text contents can not be retrieved with Get Text or with Get Cell Text

**To Reproduce**
In Salesforce
Create Account record
Create a Case under the Account
Open Accounts Cases related list view

Text from Contact Name column can not be retrieved even if visibility check has been set to false with
`Set Config    Visibility    False`

Script for reproduction
```
*** Settings ***
Library    QForce
Suite Setup    Suite Setup Actions

*** Variables ***
${username}         set this
${domain}           set this
${login_url}        set this

*** Keywords ***
Suite Setup Actions
    Open Browser    ${login_url}    chrome
    Type Text    Username    ${username}
    Type Text    Password    ${sf_pw}
    Click Text    Log In

    Verify Text    Home

*** Test Cases ***
Slot - Contact Name
    [Documentation]    Slot with text only, subject and owner are similar
    # set the domain and replace the account id with a valid one
    GoTo    ${login_url}/lightning/r/Account/0017Q00000j4QGOQA2/related/Cases/view
    Set Config    Visibility    True
    TRY
        ${con1}=    Get Text    Jill  # not found
    EXCEPT
        No Operation
    END
    ${con2}=    GetCellText    r2/c4    # empty str
    

    Set Config    Visibility    False
    ${con3}=    Get Text    Jill  # empty str
    ${con4}=    GetCellText    r2/c4    # empty str

    Set Config    Visibility    True
    TRY
        ClickText    Jill    # empty <slot>
    EXCEPT
        No Operation
    END
    Set Config  Visibility  False
    ClickText    Jill
```

PR #142, opened 2024-04-03
Tko april 2024

tkoukkari:
**Fixes:**
- RunBlock fails with uninformative error message
- GetCellText and GetText not able to get text directly under `<slot>`
- #141 
  -   Anchor coordinates were in some cases taken from parent, not from the element actually containing anchor text
- New typing issue in selenium 4.19
- Bumped Pillow to a version having latest security fixes

### DIFF
--- a/QWeb/internal/actions.py
+++ b/QWeb/internal/actions.py
@@ -388,7 +388,11 @@ def text_appearance(text: str, **kwargs: Any) -> bool:
 
 @decorators.timeout_decorator_for_actions
 def get_element_text(web_element: WebElement, expected=None, timeout: int = 0) -> Union[bool, str]:  # pylint: disable=unused-argument
-    real_text = web_element.text.strip() or web_element.get_attribute('innerText').strip()
+    real_text = (
+        (web_element.text or "").strip()
+        or (web_element.get_attribute('innerText') or "").strip()
+    )
+
     if expected is not None:
         try:
             return _compare_texts(real_text, expected.strip(), timeout)
--- a/QWeb/internal/actions.py
+++ b/QWeb/internal/actions.py
@@ -388,7 +388,7 @@ def text_appearance(text: str, **kwargs: Any) -> bool:
 
 @decorators.timeout_decorator_for_actions
 def get_element_text(web_element: WebElement, expected=None, timeout: int = 0) -> Union[bool, str]:  # pylint: disable=unused-argument
-    real_text = web_element.text.strip()
+    real_text = web_element.text.strip() or web_element.get_attribute('innerText').strip()
     if expected is not None:
         try:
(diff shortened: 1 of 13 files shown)


QWeb/internal/actions.py
```diff
@@ -388,7 +388,11 @@ def text_appearance(text: str, **kwargs: Any) -> bool:
 
 @decorators.timeout_decorator_for_actions
 def get_element_text(web_element: WebElement, expected=None, timeout: int = 0) -> Union[bool, str]:  # pylint: disable=unused-argument
-    real_text = web_element.text.strip()
+    real_text = (
+        (web_element.text or "").strip()
+        or (web_element.get_attribute('innerText') or "").strip()
+    )
```
turunenm, line 394: Maybe add comment for future readers that `web_element.text` and `web_element.get_attribute('innerText')` can be `None`
Took me a while to understand what is the purpose of empty string ternaries

QWeb/keywords/text.py
```diff
@@ -862,23 +862,28 @@ def get_text(
...
     elif '//' not in locator:
-        web_element = internal_text.get_text_using_anchor(locator, anchor)
+        # css=False means we include non-clickable texts
+        web_element = internal_text.get_text_using_anchor(locator, anchor, css=False, **kwargs)
     else:
         web_element = element.get_unique_element_by_xpath(locator, index=anchor_int, **kwargs)
-    text = web_element.text
+    # selenium's .text does not see all texts directly under slots
+    # use innerText as a backup
+    text = web_element.text or web_element.get_attribute('innerText')
```
turunenm, line 886: Probably should add `or ""` ternary because text could become `None` if `web_element.text == ""` and there's no attribute `innerText`
If the text becomes `None` then the `get_substring()` at the end of function will make `None` to show up as elements text

```
text = "" or None
text is None
```
->True

```
text = "" or None or ""
text is None
```
-> False

`web_element.text or web_element.get_attribute('innerText') or ""`

turunenm (changes requested):
One commenting suggestion
One change request

QWeb/internal/actions.py
tkoukkari: Yep, I did not get it either first but this was changed due to typing catching it.

QWeb/keywords/text.py
tkoukkari: Thanks, good catch. Will resolve with a new commit

QWeb/internal/actions.py
tkoukkari: Added comment in a new commit

QWeb/keywords/text.py
tkoukkari: It seems updates to this PR are no longer working (were done while GitHub had problems).  I will close this one and open another one

PR #143, opened 2024-04-05
Tko april 2024

tkoukkari:
Re-open of previous PR that was broken while GitHub was down for a while....

**Fixes:**
- RunBlock fails with uninformative error message
- GetCellText and GetText not able to get text directly under `<slot>`
- #141 
  -   Anchor coordinates were in some cases taken from parent, not from the element actually containing anchor text
- New typing issue in selenium 4.19
- Bumped Pillow to a version having latest security fixes


turunenm (approved):
lgtm

irosengr (approved):

Issue #145, opened 2024-04-12
Update docstring for VerifyAll keyword


turunenm:
Examples section of the kw documentation displays use of kw `VerifyTexts` instead of `VerifyAll`

Issue #144, opened 2024-04-11
Inconsistent behavior in GetInputValue vs VerifyInputValue


tkoukkari:
**Describe the bug**
Linebreaks in textarea are removed with GetInputValue but not with VerifyInputValue

**To Reproduce**
1. Go to https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_textarea_form
2. Type an address containing line breaks to the textarea, for example:
```
123 Main Street
Apt. 4B
Springfield, IL 62704
```

3. Get the input value: 
`${text}=    GetInputValue   comment`

Text will be returned without linebreaks, i.e.
`123 Main StreetApt. 4BSpringfield, IL 62704`

4. Verify the input value with the text returned:
`VerifyInputValue    comment    ${text}`

This will not work, since GetInputValue automatically removes newlines (\n), but VerifyInputValue expects them.

Adding linebreaks back will work:
`VerifyInputValue    comment    123 Main Street\nApt. 4B\nSpringfield, IL 62704`

**Expected behavior**
Value returned by GetInputValue should be the "real" value which can be used in verification directly. Removing line breaks can be easily done outside of this keyword. If it's really needed to be part of this keyword, then it probably should be controlled by an argument.

tkoukkari:
Actually both already strip newlines, so issues only arise if GetInputValue takes the value and it's used somewhere else. Adding possibility to preserver newlines via argument.

tkoukkari:
Fix included in version 3.3.0

PR #148, opened 2024-04-30
Tko may 2024

tkoukkari:
- fix for #145
- refactoring unit tests, removing deprecated method names
- refactored the way we get version number on OpenBrowser (pkg_resources is deprecated)
- fix: VerifyTable to not break on "special characters" like "[", "]", "*" or "?"
- feat: add support for treating color differences as meaningful in *Icon keywords. Also expose "tolerance" argument.
- feat: browser and driver management support (Selenium Manager) for edge and firefox
- fix: added possibility to strip newlines to GetInputValue (#144)
- fix: UploadFile to find inputs of type file when searching with xpath locator


QWeb/internal/actions.py
```diff
@@ -413,18 +416,20 @@ def get_element_text(web_element: WebElement, expected=None, timeout: int = 0) -
...
     raise QWebValueMismatchError("Text not found")
 
 
-def _compare_texts(text_to_compare: str, expected: str, timeout: int) -> bool:  # pylint: disable=unused-argument
-    if fnmatch.fnmatch(text_to_compare, expected) is False:
-        raise QWebValueMismatchError(f"Expected {expected}, found {text_to_compare}")
-    return True
+def _compare_texts(text_to_compare: str, expected: str, partial_match: bool = True) -> bool:
+    if (partial_match and expected in text_to_compare) or (
+        not partial_match and text_to_compare == expected
```
turunenm: Isn't the check for `not partial_match` redundant

This should be equivalent check
`if text_to_compare == expected or (partial_match and expected in text_to_compare)`

QWeb/internal/icon.py
```diff
@@ -283,6 +283,7 @@ def image_location(
         draw: int = 1,
         template_res_w: int = 1440,
         device_res_w: int = 1080,
+        grayscale: bool = False,
```
turunenm: Since the check used to be grayscale only I think this should default to `True`

QWeb/internal/icon.py
```diff
@@ -295,16 +296,21 @@ def image_location(
         print("*INFO* _image_location Starts")
 
         image = cv2.imread(haystack)
-        image_gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        image_gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY) if grayscale else image
```
turunenm: `image_gray` is misleading variable name as it can now be in color

Rename the var to something like `image_haystack`

QWeb/internal/util.py
```diff
@@ -156,11 +156,11 @@ def highlight_validation(color: str) -> str:
         "navy",
         "aqua",
     ]:
-        raise ValueError("Not a supported highligt color")
+        raise ValueError("Not a supported highlight color")
     return color
 
 
-def get_substring(text: str, **kwargs) -> Union[int, float, str]:
+def get_substring(text: str, strip_newlines: bool = True, **kwargs) -> Union[int, float, str]:
```
turunenm: The new flag removes all newlines instead of stripping them and should be renamed to `remove_newlines`
Stripping in Python means removing provided characters from the start/end of the string.

Semantic nitpicking but this could cause confusion at some point.

QWeb/keywords/input_.py
```diff
@@ -590,7 +590,12 @@ def verify_input_element(
 @keyword(tags=("Input", "Getters"))
 @decorators.timeout_decorator
 def get_input_value(
-    locator: str, anchor: str = "1", timeout: Union[int, float, str] = 0, index: int = 1, **kwargs
+    locator: str,
+    anchor: str = "1",
+    timeout: Union[int, float, str] = 0,
+    index: int = 1,
+    strip_newlines: bool = True,
```
turunenm: The new flag removes all newlines instead of stripping them and should be renamed to `remove_newlines`
Stripping in Python means removing provided characters from the start/end of the string.

Semantic nitpicking but this could cause confusion at some point.

QWeb/keywords/input_.py
```diff
@@ -629,6 +634,8 @@ def get_input_value(
     index : int
         If table cell contains more than one input elements or if there is some kind of
         nested structure inside of given input index may needed. Default = 1 (first)
+    strip_newlines : bool
+        Strip new lines (\n)from string, especially from multirow text areas. Default = True
```
turunenm: `remove_newlines`

QWeb/keywords/input_.py
```diff
@@ -652,7 +659,7 @@ def get_input_value(
         locator, anchor, timeout=timeout, index=index, **kwargs
     )
     val = actions.input_value(input_element, timeout=timeout, **kwargs)
-    return util.get_substring(val, **kwargs)
+    return util.get_substring(val, strip_newlines=strip_newlines, **kwargs)
```
turunenm: `remove_newlines`

QWeb/internal/util.py
```diff
@@ -176,8 +176,9 @@ def get_substring(text: str, **kwargs) -> Union[int, float, str]:
         start = end - int(kwargs.get("from_end"))  # type: ignore[arg-type]
     logger.debug("substring start: {}".format(start))
     logger.debug("substring end: {}".format(end))
-    text = str(text[start:end]).strip().replace("\n", "")
-    text = text.replace("\r", "")
+    if strip_newlines:
```
turunenm: `remove_newlines`

ruff.toml
```diff
@@ -67,7 +67,7 @@ line-ending = "auto"
 #
 # This is currently disabled by default, but it is planned for this
 # to be opt-out in the future.
-docstring-code-format = false
+# docstring-code-format = false
```
turunenm, line 70: Does this affect generated kw docs in any way after running `ruff format`?

turunenm (changes requested):
Requested some naming changes but otherwise looks good

QWeb/internal/actions.py
tkoukkari: Yes, that's true. Will change with new commit.

QWeb/internal/icon.py
tkoukkari: will update with new commit

QWeb/internal/icon.py
tkoukkari: will rename in new commit

QWeb/internal/util.py
tkoukkari: ok, didn't think of it that way but valid point. Will update with new commit.

QWeb/keywords/input_.py
tkoukkari: ok, didn't think of it that way but valid point. Will update with new commit.

QWeb/keywords/input_.py
tkoukkari: will update with new commit

QWeb/internal/util.py
tkoukkari: will update with new commit

ruff.toml
tkoukkari: not sure, but I don't think so. At least kw docs look normal

tkoukkari:
> Requested some naming changes but otherwise looks good

All of these should be now included with the latest commit

turunenm (approved):
lgtm

tkoukkari:
**Note:** MacOS/Safari tests are failing due to restriction of running local html files. This will be fixed in another PR.

PR #149, opened 2024-05-14
Mt april 2024

turunenm:
Fix to element_type handling in for some keywords.
Invalid element types were not rejected and caused unexpected behavior.

Also updated VerifyAll docstring but it should already be addressed in master


tkoukkari (approved):
lgtm

PR #151, opened 2024-05-15
http serverless execution of acceptance tests

turunenm:
When running tests outside of gh actions or local duty, it's easier to use the http files directly from the file system instead of an http server.

Added `__init__.robot` fle into __tests/acceptance__ directory which checks a global variable `${ON_HTTP_SERVER}` (defaults to `False`) which tells if the tests will be run on http server and then sets a base URI accordingly for the test cases


tkoukkari (commented):
ok

tkoukkari (approved):
ok

Issue #155, opened 2024-05-29
Table cell index notation differences


turunenm:
Targeting table cells can be done with two different notations which differ in how they behave.
One style disregards header elements `th` and one does not

excludes headers
`ClickCell    r2c4`

includes headers
`ClickCell    r3/c4` 

This should be either documented as it is now
Or modify the notation behaviors so that they work the same and that the header inclusion can be controlled with an argument

tkoukkari:
Improved documentation related to this in release 3.2.2

PR #156, opened 2024-06-17
Tko june 2024

tkoukkari:
**Changes:**
* Support non-breakable spaces when finding text directly inside slot
* Make GetTableRow "//last" argument case insensitive
* fix: partial_match=False was not forwarded correctly as boolean in few places
* fix: partial_match was not taken into account on ClickCheckbox even if it was documented to do so
* Docs: Improved table keyword's docs regarding coordinate format (#155 )
* Docs: Clarified when column headers are omitted (table keywords)
* Deps: Allow more recent versions of pyobjc
* Deps: Added numpy as a direct dependency, even if it's really a dependency to opencv. Numpy has a new release (2.x) out and there is not yet opencv release which would support the new major version.
* Deps: Bumped minimum allowed version of requests due to security alert


QWeb/internal/text.py
```diff
@@ -443,13 +443,17 @@ def get_clickable_elements(
     Returns:
         Optional[List[WebElement]]: List of clickable WebElement objects, or None if none found.
     """
-    partial = kwargs["partial_match"]
+    partial = util.par2bool(kwargs.get("partial_match", False))
```
turunenm, line 446: I think it's better to get the default with `CONFIG["PartialMatch"]`

QWeb/internal/table.py
```diff
@@ -377,8 +378,8 @@ def get_columns(self) -> list[WebElement]:
                         // If the table doesn't have a header, count the columns in the first body row
                         var firstBodyRow = tableRef.querySelector('tbody tr');
                         if (firstBodyRow) {
-                            headers = firstBodyRow.querySelectorAll('th');
                         }
+                            headers = firstBodyRow.querySelectorAll('th');
```
turunenm: I don't understand this.
Isn't the if block now empty and useless?

turunenm (commented):
Two minor remarks, lgtm otherwise

QWeb/internal/text.py
tkoukkari: Did it like this, as it's already handled in the calling function:
```
def get_all_text_elements
...
kwargs["partial_match"] = kwargs.get("partial_match", CONFIG["PartialMatch"])
```

..so kwargs{"partial_match"] should always have a value.

I think this is one place where we should make partial_match a real argument (not kwargs), but let's handle that when changing other kwargs...

QWeb/internal/table.py
tkoukkari: hmm, I do not remember changing that one...?

turunenm (approved):
lgtm

PR #157, opened 2024-06-27
fix: incorrectly forwarding kwargs to get_checkbox_by_locator()

tkoukkari:
Not sure why this wasn't spotted, but kwargs got incorrectly forwared to get_checkbox_by_locator()...this will cause an error in specific situation when using ClickCheckbox


turunenm (approved):
lgtm

samijt (approved):

PR #161, opened 2024-08-14
Tko support selenium grid

tkoukkari:
* Added Selenium grid / remote browser support for Chrome, Firefox, Edge and Safari
* Refactored Chrome's open browser due to complexity


QWeb/internal/browser/chrome.py
```diff
@@ -60,79 +60,102 @@ def open_browser(
...
+
+    if user.is_root():
+        options.add_argument("no-sandbox")
+
+    return options
+
+
+def resolve_chromedriver_path(executable_path: str) -> str:
+    """Resolve the path to the ChromeDriver executable."""
+    return util.get_rfw_variable_value("${CHROMEDRIVER_PATH}") or executable_path
```
turunenm: Add debug log which tells the origin of the path (rfw var / arg executable_path) and the value of the path.

I couldn't find `${CHROMEDRIVER_PATH}` from QWeb documentation.
Is it only for internal use?

turunenm (commented):
The purposes of all variables that are retrieved from RFW instance should be documented at least in the Python sources.
Silent argument value overrides with RFW vars should be debug/trace logged.

Implementation is good to go

QWeb/internal/browser/chrome.py
tkoukkari: tbh no idea of all these variables, old code. Apparently you could give chromedriver path as variable, but afaik with selenium manager this probably is not needed unless you specifically want to use some other chromium based browser or for example portable version of the browser. That part is documented, just not with the rfw variable:

# Use Chromium instead of Chrome:
OpenBrowser    http://google.com     chrome    chrome_path=/path/to/chromium/chrome.exe
OpenBrowser    http://google.com     chrome    executable_path=/path/to/my/chromedriver.exe

I will add the debug log part, but I think we should sort out all of these variables and document them,, but let's handle those in another branch/PR.

tkoukkari: actually even logging this is not that important, as in most cases it will be empty (default), meaning that chromedriver is either on PATH or handled by selenium manager. On the other hand, if custom path is used, it will be either in this variable or as open browser argument -> not much additional value of logging it.

@turunenm If you really want this logging to be added, I will log it but only if custom value is used.

tkoukkari: added logging if custom chromedriver path is given in new commit

turunenm: Please add the logging because if someone 'accidentally' declares one of the overriding rf variables in their test script it'll be really difficult to figure out why qweb is behaving unexpectedly.

QWeb/internal/browser/chrome.py
```diff
@@ -127,7 +126,10 @@ def create_chrome_options(chrome_args: Optional[list[str]], **kwargs: Any) -> Op
 
 def resolve_chromedriver_path(executable_path: str) -> str:
     """Resolve the path to the ChromeDriver executable."""
-    return util.get_rfw_variable_value("${CHROMEDRIVER_PATH}") or executable_path
+    driver_path = util.get_rfw_variable_value("${CHROMEDRIVER_PATH}") or executable_path
+    if driver_path:
+        logger.debug(f"Chromedriver path: {driver_path}")
```
turunenm, line 131: Setting path with `executable_path` should be intentional by user and IMO not necessary to log.
But the override with with rf variable is a 'global' side-effect and context about the change should be provided in the log message.

```
rf_path = util.get_rfw_variable_value("${CHROMEDRIVER_PATH}")
driver_path = rf_path or executable_path
if driver_path and rf_path:
        logger.debug(f"Setting webdriver executable path with Robot Framework variable ${{CHROMEDRIVER_PATH}} to ${driver_path}")
```

_internal/browser/edge.py_ has similar override with `${EDGEDRIVER_PATH}` and that override should also be logged in this fashion

turunenm (changes requested):
Logging the silently overridden `driver_path` with `${CHROME/EDGEDRIVER_PATH}` should provide contextual info about what has happened.

QWeb/internal/browser/chrome.py
tkoukkari: nope, sorry this is how it will be. This will log the one that will be used, regardless which one was intentionally or unintentionally changed.

tkoukkari: edgedriver custom path logging added

turunenm (approved):

PR #163, opened 2024-09-09
feat: add the same workaround for getting texts directly in slots to …

tkoukkari:
…VerifyText as we already have for ClickText.

Basically slots are considered invisible by default, since they do not have offsets (which is one of the checks we do). Since Salesforce seems to commonly use the practice of having texts directly in slots, we will take parent element from slots that have direct texts and are under <a> tag.


QWeb/internal/icon.py
```diff
@@ -384,6 +385,11 @@ def image_location(
                         cv2.imwrite("temp_matched_area.png", image)
                     break
 
+        # No match found, set to a default scale ratio and original image
+        if best_highest_max_val == 0.0:
+            best_scale_ratio = 1.00
+            best_matched_image = template
```
turunenm, line 378: Add debug message that no match was found

QWeb/internal/text.py
```diff
@@ -104,13 +105,23 @@ def get_text_elements(text: str, **kwargs) -> Optional[list[WebElement]]:
             web_elements = _get_contains_text_element(text, **kwargs)
         except NoSuchFrameException:
             logger.debug("Got no such frame from contains text")
+    # find text directly from slots with supported parent elements
+    xpath_elements = None
```
turunenm: Initialize the var with `None` on L97 where it's defined and remove this line

QWeb/internal/text.py
```diff
@@ -104,13 +105,23 @@ def get_text_elements(text: str, **kwargs) -> Optional[list[WebElement]]:
...
+    slot_elements = get_slot_elements(text, partial)
```
turunenm, line 110: Define and initialize `slot_elements` at the top of the function for consistency

QWeb/internal/text.py
```diff
@@ -104,13 +105,23 @@ def get_text_elements(text: str, **kwargs) -> Optional[list[WebElement]]:
...
+    if slot_elements:
+        xpath_elements = element.get_visible_elements_from_elements(slot_elements, **kwargs) or None
+    if xpath_elements:
+        if web_elements is None:
```
turunenm, line 114: I believe this can cause `UnboundLocalError` if both L99 and L105 raise an error.
On L96 where  `web_elements` is defined it should be initialized with `None`

QWeb/internal/table.py
```diff
@@ -166,7 +166,7 @@ def get_using_text_in_coordinates(self, coordinates: str, anchor: str, **kwargs)
         column: Optional[int]
         row_elem = None
         cell = None
-        locator = coordinates.split("/")
+        locator = coordinates.rsplit("/", 1)
```
turunenm: While this solves the immediate need for using '/' (f-slash) chars in text finders for cells this is not a proper solution.

Consider this kind of a table
```
|       | 2022/2023 | 2023/2024 |
---------------------------------
| 13/09 |    A1     |     B1    |
---------------------------------
| 24/12 |    A2     |     B2    |
```

If user wants to search for columns which have f-slashes then the user needs to know and understand that the r?X/c?Y can be flipped from the locator.
`Click Cell    c?13/09/2024/r2`

User could also use f-slash searches for both row and col which won't work with this
`Click Cell    r?2023/2024/c?13/09`


F-slash handling should be more robust and probably requires having an escape character and parsing rules for it.

turunenm (changes requested):

QWeb/internal/text.py
tkoukkari: not sure since called functions return None if nothing is found. Maybe if some exception occurs but those should be catched anyway. But I'll add it "just in case"

QWeb/internal/table.py
tkoukkari: `Click Cell    r?2023/2024/c?13/09`

I don't think this is accepted format anyway? Why would they be finding via text search knowing both row and column? This is not in any examples in our documentation afaik. It's either r? or c?, right?

But true, I did not take into account that column can be given first...or can it? Never seen it used like this.

tkoukkari: Tested briefly, and as far as I can tell the other way around has never worked (Click Cell    c?13/09/2024/r2) and also using text search in both row and column has never been supported. This would be a new feature and out of the scope of this PR.

QWeb/internal/icon.py
tkoukkari: handled with new commit

QWeb/internal/text.py
tkoukkari: handled with new commit

QWeb/internal/text.py
tkoukkari: handled with new commit

QWeb/internal/text.py
tkoukkari: handled with new commit

QWeb/internal/table.py
tkoukkari: meaning it has never worked even for values without "/" in them, for example:
`ClickCell            c3/r?Jim`

turunenm: Getting the cross section cell with text searches is a valid scenario if you don't want to hard code indices or if the ordering of the table is not guaranteed.

Using text search on both row and col seems to work at least with this trivial example at _table.html_
```
UseTable      Sample
GetCellText  r?Jill/c?Last
--> 'Smith'
```

image

I misremembered that you would be able to flip the cells but that alleviates the problem with this solution proposal, you could only look for rows which contain f-slashes and not columns.


Lets deal with f-slashes in a separate PR?

tkoukkari: Wouldn't affect the example given above, but removed with new commit. Added also note to kw docs that "/" is not supported.

turunenm (approved):
Good to go

samijt (approved):

PR #169, opened 2024-11-15
fix: fix issue of inputs not being found by attribute (=javascript) search

tkoukkari:
…in some salesforce views

...as they are under pseudoelement which cannot be parsed via javascript. We will do xpath search as a backup if nothign is found with "normal" means

### DIFF
--- a/QWeb/internal/element.py
+++ b/QWeb/internal/element.py
@@ -642,22 +642,26 @@ def get_elements_by_attributes(
     if "tag" in kwargs:
         css = str(kwargs.get("tag"))
     try:
-        elements = javascript.get_all_elements(css)
-        if any_element:
-            return elements
-        matches = javascript.get_by_attributes(
-            elements,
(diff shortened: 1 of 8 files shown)


QWeb/internal/element.py
```diff
@@ -642,22 +642,26 @@ def get_elements_by_attributes(
...
+            matches = {"full": [], "partial": []}
 
-        # try with xpath if no matches
-        # there have been few where css search doesn't work
-        # this is supported only if tag argument is given
+        # try with xpath if no matches and tag is givenor if css contains xpath
+        # there have been few cases where css search doesn't work
         if len(matches.get("full", [])) == 0 and len(matches.get("partial", [])) == 0:
             try:
+                css = css if xpath_search else f"//{css}"
```
turunenm: I would name var this as `xpath` to make it's purpose more obvious

tkoukkari: ok, yes true. Will update with a new commit

turunenm (approved):

PR #172, opened 2025-02-03
feat: added new keywords MouseDown, MouseUp, MouseMove and ClickCoord…

tkoukkari:
…inates

...also added test for RightClick


samijt (approved):

turunenm (approved):
Might be worth adding to `mouse_move` and `click_coordinates` doc strings that 0,0 is the top-left corner of the viewport and the coord values grow when moving towards bottom right corner. Suggesting because school math coords start from bottom-left and grow towards top-right and the difference might throw some people off.

Also it might be a good idea to add `mouse_get_coordinates` helper kw to avoid the hassle of using javascript to get the mouse coords.

Code looks good as it is 👍

PR #173, opened 2025-02-05
Tko get coordinates

tkoukkari:
* Added **GetCoordinates** keyword which handles scrolling if possible and can take overlay/top bar into account with an argument.
* Also added **overlay_height** argument to **ScrollText**
* Improved documentation regarding linux installation + note about preferring to use automatic driver management via Selenium Manager


QWeb/internal/actions.py
```diff
@@ -136,6 +136,22 @@ def scroll(web_element: WebElement, timeout: int) -> None:  # pylint: disable=un
...
+
+    # Keep checking until the scroll position stops changing
+    # This was needed in firefox because the scroll position was not updated immediately
+    while previous_scroll != current_scroll:
+        previous_scroll = current_scroll
+        time.sleep(0.2)  # Small delay to allow scrolling to stabilize
+        current_scroll = driver.execute_script("return window.pageYOffset;")
+
+    # scroll up by overlay height
+    driver.execute_script(f"window.scrollBy(0, -{overlay_height})")
```
turunenm: Does this work with negative values for cases where there's an overlay footer instead of a header?

QWeb/internal/actions.py
```diff
@@ -136,6 +136,22 @@ def scroll(web_element: WebElement, timeout: int) -> None:  # pylint: disable=un
     javascript.execute_javascript("arguments[0].scrollIntoView();", web_element)
 
 
+def scroll_overlay_adjustment(overlay_height: int = 0) -> None:  # pylint: disable=unused-argument
```
turunenm: `overlay_height` -> `overlay_offset`?
the overlay could be a footer and providing a negative value should take care of the positioning.
Might be kind of annoying to document as scroll direction is not that intuitive (at least for me)

QWeb/keywords/element.py
```diff
@@ -761,3 +761,132 @@ def verify_attribute(
...
     except QWebValueError as e:
         raise e
+
+
+@keyword(tags=["Getters"])
+def get_coordinates(
+    locator: Union[WebElement, str],
+    anchor: str = "1",
+    element_type: Optional[str] = None,
+    overlay_height: int = 0,
```
turunenm: `overlay_height` -> `overlay_offset`?

QWeb/keywords/element.py
```diff
@@ -761,3 +761,132 @@ def verify_attribute(
...
+    locator : str
+        Visible text, attribute value Xpath expression or WebElement.
+        See GetWebElement for more information.
+    anchor : int
+        Used when element_type is defined. Default=1 (first match)
+    element_type : string
+        Define element type/preferred searching method
+        (available types: text, input, checkbox, item, dropdown or css).
+    overlay_height: int
+        Height of the overlay to be added to the element's height. Default=0
```
turunenm: `overlay_height` -> `overlay_offset`?

QWeb/keywords/text.py
```diff
@@ -1261,12 +1262,23 @@ def scroll_text(
         ScrollText       Address
         ScrollText       Address    Billing
 
+        # Scroll to element with text "Address" and adjust overlay height
+        ScrollText       Address    overlay_height=30
+
+    Accepted kwargs:
+        overlay_height: Amount of pixels to adjust overlay height in cases where there is
+        a sticky header that covers the text.
```
turunenm: `overlay_height` -> `overlay_offset`?

turunenm (commented):
Code looks good.

Consider changing 
`overlay_height` -> `overlay_offset`
as the overlay element could be a footer and providing a negative value should take care of the positioning.

QWeb/internal/actions.py
tkoukkari: not sure, tbh did not think about the footer as I've only seen header / top bar overlays. Will check

QWeb/internal/actions.py
tkoukkari: Yes, makes sense...I was only thinkin about header. I will check footer, need to generate a test for that too

tkoukkari: turns out it's close to impossible to get element under footer as element is always scrolled to the top. For now I still added support for negative numbers but still pondering if it would be simpler to just support top overlay / positive numbers.

QWeb/keywords/text.py
tkoukkari: done

QWeb/keywords/element.py
tkoukkari: done

QWeb/keywords/element.py
tkoukkari: done

QWeb/internal/actions.py
tkoukkari: I left possibility to have negative offset to helper function in case this is needed in the future. For now current keyword always scrolls the element to top, so it can't be under footer / bottom overlay. Made getcoordinates only approve positive offset values

turunenm (approved):

PR #174, opened 2025-02-07
refactor: added support for in-place mouse clicks (with None as locator)

tkoukkari:
Also changed coordinates to float + added documentation.


QWeb/internal/actions.py
```diff
@@ -204,19 +204,27 @@ def right_click(element: WebElement) -> None:
...
 
-def mouse_down(element: WebElement) -> None:
+def mouse_down(element: Optional[WebElement] = None) -> None:
     driver = browser.get_current_browser()
     ac = ActionChains(driver)
-    ac.click_and_hold(element).perform()
+    if element:
+        ac.click_and_hold(element)
+    else:
+        ac.click_and_hold()
```
turunenm: if-else is not necessary here

`ac.click_and_hold(None)` is equivalent to `ac.click_and_hold()`

If there's a need for differing error messages for `WebElement` / `None` in the future then keeping the if-else could be beneficial

QWeb/internal/actions.py
```diff
@@ -204,19 +204,27 @@ def right_click(element: WebElement) -> None:
...
 
-def mouse_up(element: WebElement) -> None:
+def mouse_up(element: Optional[WebElement] = None) -> None:
     driver = browser.get_current_browser()
     ac = ActionChains(driver)
-    ac.release(element).perform()
+    if element:
+        ac.release(element)
+    else:
+        ac.release()
```
turunenm: if-else is not necessary here

`ac.click_and_hold(None)` is the equivalent to `ac.click_and_hold()`

If there's a need for differing error messages for `WebElement` / `None` in the future then keeping the if-else could be beneficial

turunenm (commented):
Approving after comments are resolved

QWeb/internal/actions.py
tkoukkari: ah, true. Will fix.

turunenm (approved):

QWeb/internal/actions.py
tkoukkari: Fixed with a new commmit

QWeb/internal/actions.py
tkoukkari: Fixed with a new commmit

PR #175, opened 2025-03-05
Switch to Openbox window manager in linux pipeline

tkoukkari:
Switching to Openbox window manager due to inability to maximize/minimize or set size with matchbox.

Some intentional runs with matchbox to verify the problem and then changed to Openbox.

### DIFF
--- a/QWeb/internal/actions.py
+++ b/QWeb/internal/actions.py
@@ -34,6 +34,7 @@ from typing import Any, Optional, Union
 from robot.api import logger
 from selenium.common.exceptions import (
     ElementNotInteractableException,
+    JavascriptException,
     MoveTargetOutOfBoundsException,
     NoSuchElementException,
     WebDriverException,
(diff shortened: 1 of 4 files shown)


QWeb/internal/actions.py
```diff
@@ -151,7 +151,19 @@ def scroll_overlay_adjustment(overlay_offset: int = 0) -> None:  # pylint: disab
...
-    driver.execute_script(f"window.scrollBy(0, {scroll_amount})")
+    if driver.capabilities["browserName"].lower() in browser.firefox.NAMES:
+        # Firefox 135+ needs these settings in some Windows environments
+        driver.execute_script("document.body.style.willChange = 'scroll-position';")
+        driver.execute_script("document.body.style.overflowAnchor = 'none';")
+        driver.execute_script(f"window.scrollBy(0, {scroll_amount})")
+
+        # Restore default behavior (to prevent unexpected side effects)
+        driver.execute_script("document.body.style.willChange = '';")
+        driver.execute_script("document.body.style.overflowAnchor = '';")
```
turunenm: I'd wrap these in try-except-finally so that in case if something goes sour there's better chance to preserve the original style of the body

Something like
```
        try:
            driver.execute_script("document.body.style.willChange = 'scroll-position';")
            driver.execute_script("document.body.style.overflowAnchor = 'none';")
            driver.execute_script(f"window.scrollBy(0, {scroll_amount})")
        except Exception as e:
            pass
        finally:
            # Restore default behavior (to prevent unexpected side effects)
            driver.execute_script("document.body.style.willChange = '';")
            driver.execute_script("document.body.style.overflowAnchor = '';")
```

turunenm (commented):
lgtm
check the comment and I'll approve after that

QWeb/internal/actions.py
turunenm: I think it would also be good to first read the style values which are changed so that they will be restored to the originals instead of the current fixed values

tkoukkari: yep, makes sense. I also realized that instead of finally setting them to empty we probably should put the previous values back. Both seem to be "auto" by default for me, but not sure if that's universal. Maybe better to read the value and put the same one back

tkoukkari: changed with new commit

turunenm (approved):

PR #176, opened 2025-03-20
fix: Handle full match in table keywords in a different way, which su…

tkoukkari:
…pports getting full match from known attributes containing column header (if available) in addition to text.

At the same go refactored def get_col_header and verify_col_header -> moved internal support functions to /internal/table. Also slightly renamed them and added own func for getting the column index.

Also changed PDF reading functionality from Slate3k to pypdf.

### DIFF
--- a/QWeb/internal/table.py
+++ b/QWeb/internal/table.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 # ---------------------------
 from __future__ import annotations
-from typing import Optional, Union
+from typing import List, Optional, Union
 
 import fnmatch
 import re
@@ -167,6 +167,7 @@ class Table:
         row_elem = None
         cell = None
         locator = coordinates.split("/")
+        partial_match = util.par2bool(kwargs.get("partial_match", CONFIG["PartialMatch"]))
         if locator[0].startswith("r?"):
             row_elem = self.get_row(locator[0][2:], anchor)
             # take last row as "row" if row_elem is not None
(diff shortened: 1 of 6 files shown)


QWeb/internal/table.py
```diff
@@ -386,3 +393,22 @@ def get_columns(self) -> list[WebElement]:
...
+            index = column_texts.index(locator)
+            index = index + 1
+        except ValueError as e:
+            raise QWebValueError(f"Matching column not found for locator {locator}.") from e
+        return index
+
+    def get_column_header_texts(self, columns: List[WebElement]) -> List[str]:
+        # prefer using title, then aria-label, then text
+        column_texts = [
+            c.get_attribute("aria-label") or c.get_attribute("title") or c.text for c in columns
```
mturunen-copado: Comment does not match the behavior, order of preference is `aria-label, title, text`

QWeb/internal/table.py
```diff
@@ -386,3 +393,22 @@ def get_columns(self) -> list[WebElement]:
                     }
                 return(columns(arguments[0]));"""
         return javascript.execute_javascript(js, self.table)
+
+    def get_full_match_column_index(self, locator: str) -> int:
+        """Get the index of the column based on the locator text. Full match"""
```
mturunen-copado: Maybe add into docstring that this returns 1-based index instead of usual 0-based index

mturunen-copado (commented):
lgtm
check the comments and I'll approve after that

QWeb/internal/table.py
tkoukkari: fixed with a new commit

QWeb/internal/table.py
tkoukkari: fixed with a new commit

mturunen-copado (approved):

turunenm (approved):

PR #177, opened 2025-04-03
Tko dependency updates

tkoukkari:
* Bumped major versions of **numpy** and **Pillow**
* Bumped **pynput**, **pyperclip**, **opencv** and **scikit-image** minor versions
* Updated pipeline python version to Python 3.11


QWeb/internal/screenshot.py
```diff
@@ -147,6 +147,7 @@ def _draw_contours(diff: ndarray, ref_image_c: ndarray) -> ndarray:
     index = 1 if cv2.__version__.startswith("3") else 0  # type: ignore
 
     for c in contours[index]:
+        assert isinstance(c, ndarray)
```
turunenm: Is  `assert` intended for debugging or runtime checks?

If this is intended for active runtime check and someone builds QWeb from sources with `-O` flag then this line will be disabled and can lead to unexpected situation.

https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

turunenm: If this is for runtime check then the line should be changed to something like (mimicked from the py doc)
`if not isinstance(c, ndarray): raise AssertationError`

turunenm (commented):
Check the `assert` comment

QWeb/internal/screenshot.py
tkoukkari: ah, yes. I will fix this with a new commit. I will actually then add that notification functionality in this pr too (just to save time and test rounds), so maybe I'll mark this as a draft for a while.

tkoukkari: Ok, I don't seem to be able to put this back as draft. In any case this was there just to make typing happy, forgot to make the final change there. There was a bit of a change of types in latest cv2 version. We also still supported cv2 version 3 (very old) which had also differences. I now simplified this to make it match the signature/types of minimum cv2 expected version we require in deps:

`contours, _ = cv2.findContours(thresh.copy(), cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_SIMPLE)`

tkoukkari: fixed with a new commit

QWeb/internal/javascript.py
mturunen-copado, line 1: Would be nice to create a clean mechanism for loading these js snippets from .js files instead of having them as inline strings.
Not in scope for this PR, but just raising it up for future consideration

samijt (approved):

turunenm (approved):

QWeb/internal/screenshot.py
tkoukkari: doh, fixed all the reasons for needing that assert for testing in the first place, but forgot to touch the assert itself

turunenm (approved):

PR #190, opened 2025-09-03
Modernize wait function

### LINKED ISSUE #185: MacOs Edge support
Edgedriver was removed from base (-latest) MacOs images. Mac Edge runs were disabled from the pipeline. Add then back if/when Edge is added back to MacOs images.

tkoukkari:
Adds new, enhanced WaitStrategy and sets it as default. Old one can be taken into use by setting config "WaitStrategy" to value "legacy".

Adds the following configurable values:
* WaitStrategy
* RenderWait
* SpinnerCSS

+ refactored config parameters so that docs display them in alphabetical order, adds few unit tests and documentation for the new configs.


QWeb/internal/config_defaults.py
```diff
@@ -99,6 +99,8 @@
     "LogMatchedIcons": (False, util.par2bool),
     "ShadowDOM": (False, util.par2bool),
     "HighlightColor": ("blue", util.highlight_validation),
+    "SpinnerCSS": ("none", None),
+    "DomQuiet": (200, int)
```
turunenm: In this PRs context I kind of knew what `DomQuiet` is for but I still had to check the docstring

I'd make this name more explicit, maybe `DomUnchangedMilliseconds`
This will be more cumbersome to type but I think it's pretty safe assume that it won't be set more than once in a suite in a large majority of cases where it changing this is needed.

QWeb/internal/xhr.py
```diff
@@ -16,77 +16,250 @@
...
+from typing import Optional
 from QWeb.internal import javascript
+from QWeb.internal.exceptions import QWebDriverError
+from QWeb.keywords import config
 
 
-def setup_xhr_monitor() -> bool:
-    """Inject jQuery if needed and check if page is ready.
+# Install monitor (fetch/XMLHttpRequest + MutationObserver). Idempotent.
+JS_INSTALL_MONITOR = r"""
```
turunenm, line 27: There's quite a bit of new JS.

Would now be a good time to move most of our inline JS's into separate .js files which are loaded on the fly?

Syntax highlighting and Intellisense would be so nice to have when viewing the JS functions 🤔

turunenm (commented):
I'll continue reviewing after decision is made about moving inline JS to separate files

QWeb/internal/config_defaults.py
tkoukkari: @turunenm I completely understand where you are coming from, and to be honest I opted for a short name with documentation. We can change the name, but I feel even DomUnchangedMilliseconds feels a bit technical and is imho too long for a setting. Would you have any other ideas? I put couple of proposals for food for thought below:

1. SettleTime
2. UIStable
3. DomStable
4. UISettled
5. IdleUI
6. RenderWait
7. PageStabilize / PageStable
8. CalmPeriod

(ms. could be added to the end but personally I don't feel the need. But I'm still open if you feel we should have it there)

QWeb/internal/xhr.py
tkoukkari: dunno, that would need a lot more tedious work and testing. Would it be better to move all js code to .js files in separate branch / PR?

QWeb/internal/config_defaults.py
turunenm: I like `RenderWait` a lot as it tells what it does and why.

I'm bit on a fence about the type info in the name as having type info in name is bit clunky and will feel redundant if you know by heart what type the value should be, but on the other hand it's also annoying to check what type should go in.

Another thing for the type info is that QWeb generally deals with time in seconds and I'm fairly sure that I would try inserting float seconds into this setting unless checking it from the docs before using it.

I guess I'm slightly leaning on having the `Ms` suffix in the config name but you can make the call.

QWeb/internal/xhr.py
turunenm: Could the work be started by creating a mechanism for loading these new JS's from file?

Then once this PR is completed create a new branch for moving rest of the long JS's to separate sources.
For the rest of the work I'm guessing that it would be good starting rule of thumb that if instance of inline JS is less then 8 lines long and has no branching it can stay inline.

I understand that the request is bit annoying but reviewing such complex docstringed inline JS kind of requires creating placeholder files in order to have syntax hl and intellisense to lessen the cognitive load.

I assume that you have used such placeholders when developing the JS funcs and then have been copy-pasting the funcs into this source when trying it out.
Or there's some smarter way of developing these that I'm not aware of 😅 

And if placeholder files need to be created for development and review then why not just have them as separate files persistently?

QWeb/internal/config_defaults.py
samijt: I also prefer `RenderWait `without any notion of unit. I think we should follow[ RF time format](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#time-format) which supports also milliseconds and hence no need to tell explicitly in the config name. This woud mean that the type of the config should be string and not int and default value is then "200ms". Anyhow, if might be useful to log the set time so that the user understands what was actually set.

tkoukkari: Ok, I'll change it to "RenderWait" and make sure it understands rfw format (it doesn't atm as validate function is int).

QWeb/internal/xhr.py
tkoukkari: yes, as agreed we will do this next. The plan is to do one release with just the current changes + new wait function and have the js change on next one. I created FR #191 to track this.

QWeb/internal/util.py
```diff
@@ -402,3 +402,41 @@ def remove_stale_elements(elems: Optional[List[WebElement]]) -> Optional[List[We
...
+        if n < 0:
+            raise ValueError("Milliseconds value must be non-negative.")
+        return f"{n}ms"
+
+    raise ValueError(f"Invalid millisecond value: '{value}'")
+
+
+def parse_ms(value: str) -> int:
+    """Convert a canonical '<int>ms' string into an int."""
+    return int(value[:-2])
```
turunenm: This _looks_ dangerous, even though it's only called with a validated config value and is currently safe it would be better to handle it like the validator above does.
So if in future someone decides to use the function with something else it won't blow up.

```
    v = value.strip().lower().replace(" ", "")
    if v.endswith("ms"):
        v = v[:-2]
    try:
        n = int(v)
    except ValueError as e:
        raise ValueError(f"Invalid millisecond value: {value!r}") from e
    return n
```

You can then use this func also in the `validate_ms` to avoid code duplication

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+    xhrPatched: false,
+    lastMutationTs: Date.now(),
+    observerStarted: false
+  });
+
+  try {
+    if (!window.__xhrMon.fetchPatched && typeof window.fetch === "function") {
+      const _orig = window.fetch;
+      window.fetch = function() {
+        try { window.__xhrMon.pending++; } catch(e) {}
```
turunenm, line 54: Silent error discarding of all errors seems bit questionable.

I don't know how often and what errors occur here and if they clog the console but some handling could be beneficial.
I'm guessing that the `++` operation is quite likely to fail which throws and can be silently ignored, but other, more rare and yet unidentified, errors probably should get logged somewhere

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+        const p = _orig.apply(this, arguments);
+        const dec = function(){ try { window.__xhrMon.pending--; } catch(e) {} };
```
turunenm, line 56: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+        return p && typeof p.finally === "function"
+          ? p.finally(dec)
+          : p.then(function(r){ dec(); return r; }, function(e){ dec(); throw e; });
+      };
+      window.__xhrMon.fetchPatched = true;
+    }
+  } catch(e){}
```
turunenm: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+
+  try {
+    if (!window.__xhrMon.xhrPatched && window.XMLHttpRequest && window.XMLHttpRequest.prototype) {
+      const _open = XMLHttpRequest.prototype.open;
+      const _send = XMLHttpRequest.prototype.send;
+      XMLHttpRequest.prototype.open = function() { this.__xhrTracked = true; return _open.apply(this, arguments); };
+      XMLHttpRequest.prototype.send = function() {
+        if (this.__xhrTracked) {
+          try { window.__xhrMon.pending++; } catch(e) {}
```
turunenm, line 72: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+          this.addEventListener("loadend", function(){ try { window.__xhrMon.pending--; } catch(e) {} }, { once:true });
```
turunenm, line 73: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+
+  try {
+    if (!window.__xhrMon.observerStarted && window.MutationObserver) {
+      const obs = new MutationObserver(function(){ window.__xhrMon.lastMutationTs = Date.now(); });
+      obs.observe(document.documentElement || document.body, {
+        childList:true, subtree:true
+      });
+      window.__xhrMon.observerStarted = true;
+    }
+  } catch(e){}
```
turunenm: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+
+# Lightweight status (no spinner logic here anymore)
+JS_STATUS_LITE = r"""
+return (function (quietMs) {
+  var ready = (document.readyState === "complete");
+
+  var jqActive = null;
+  try {
+    if (typeof jQuery !== "undefined" && jQuery && typeof jQuery.active === "number") jqActive = jQuery.active;
+  } catch(e){}
```
turunenm, line 108: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+
+  var pending = 0, last = Date.now();
+  try {
+    pending = (window.__xhrMon && typeof window.__xhrMon.pending === "number") ? window.__xhrMon.pending : 0;
+    last    = (window.__xhrMon && window.__xhrMon.lastMutationTs) ? window.__xhrMon.lastMutationTs : last;
+  } catch(e){}
```
turunenm, line 114: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+  try {
+    var sels = Array.isArray(selectors) ? selectors : [];
+    for (var i=0;i<sels.length;i++){
+      var nodes = document.querySelectorAll(sels[i]);
+      for (var j=0;j<nodes.length;j++){
+        if (visible(nodes[j])) return true;
+      }
+    }
+    return false;
+  } catch (e) {
```
turunenm, line 151: Silent error discarding of all errors seems bit questionable.

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+    return false; // if probe fails, treat as not-busy (avoid blocking)
+  }
+})(arguments[0]);
+"""
 
 
 def setup_xhr_monitor() -> bool:
+    try:
+        ok = javascript.execute_javascript(JS_INSTALL_MONITOR)
+        if ok is not True:
```
turunenm: Cast the `ok` to `bool` before checking the identity or remove the cast from `return` statement if `execute_javascript` is guaranteed to return `bool`

```
>>> "true" is not True
<stdin>:1: SyntaxWarning: "is not" with a literal. Did you mean "!="?
True
```

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+        if isinstance(busy, bool):
+            return busy
+        logger.debug(f"is_spinner_busy: unexpected return {type(busy)}")
+        return None
+    except JavascriptException as e:
+        logger.debug(f"is_spinner_busy failed: {e}")
+        return None
+
+
+def _parse_spinner_selectors() -> Optional[list[str]]:
```
turunenm, line 193: Could be also written like this if you want to make the func more compact

```
def _parse_spinner_selectors() -> Optional[list[str]]:
    raw = config.get_config("SpinnerCSS")

    if isinstance(raw, str):
        # Interpret only non empty, comma-separated CSS selectors
        if raw.strip().lower() not in ("", "none", "null", "false", "off"):
            return [s.strip() for s in raw.split(",") if s.strip()]
    return None
```

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+        return [s.strip() for s in raw.split(",") if s.strip()]
+
+    return None
+
+
+def wait_xhr(timeout: float = 15.0,
+             poll_interval: float = 0.1) -> None:
+    """
+    Order: readyState -> network idle -> spinner gone -> DOM quiet (bounded).
+    - `quiet_ms`: quiet window needed to call DOM "settled"
```
turunenm: Could you add the constants and their effects on the behavior to the docstring?
Then when hovering over the function call somewhere else in an IDE you can get a better grasp of the func without traversing to this source file

QWeb/internal/xhr.py
```diff
@@ -16,12 +16,256 @@
...
+        logger.debug(f"setup_xhr_monitor failed: {e}")
+        raise QWebDriverError(e)  # pylint: disable=W0707
+
+
+def get_light_status(quiet_ms: int = 400) -> Optional[dict]:
+    """Full status dict. None on failure."""
+    try:
+        st = javascript.execute_javascript(JS_STATUS_LITE, quiet_ms)
+        if isinstance(st, dict):
+            return st
```
turunenm, line 171: Since there is a handful of known values that will be checked from the status it would be nice to use a `dataclass` with proper types instead of `dict` for passing the data.

turunenm (changes requested):
Address silent discarding of *all* errors in multiple places in the inline JS's

Converting page status properties from `dictionary` to `dataclass` could be very convenient for future troubleshooting.

QWeb/internal/xhr.py
turunenm: One option would be passing a debug flag into the functions to enable collecting and returning errors only when needed.

QWeb/internal/xhr.py
tkoukkari: would be nice but we are currently not using dataclasses anywhere so this should be fine.

QWeb/internal/xhr.py
tkoukkari: ok

tkoukkari: not even needed actually, refactored a bit

QWeb/internal/xhr.py
tkoukkari: added debug mode to the js installation script which we call once in setup_xhr_monitor.

The ones that will be retried were left as they are. This is for all "silent error" comments.

QWeb/internal/xhr.py
tkoukkari: Yes, will change that part. Also noticed that this would fail if "SpinnerCSS" config does not exist. Will catch that and update with new commit.

QWeb/internal/util.py
tkoukkari: yep, it was like that because we already validate it before and don't accept other kind of formats. But will change as suggested in case this will be re-used somewhere else in the future. Will update with a new commit.

QWeb/internal/xhr.py
tkoukkari: ok, will update with a new commit

turunenm (approved):

PR #193, opened 2025-09-30
Move javascript to js

### LINKED ISSUE #191: Move all javascript code to .js files
To enable better ide syntax highlighting and easier management/updates etc., move all javascript code to separate .js files.

tkoukkari:
Implements #191


QWeb/internal/frame.py
```diff
@@ -78,16 +78,21 @@ def wait_page_loaded() -> None:
...
             driver.switch_to.default_content()
     timeout = CONFIG["XHRTimeout"]
-    if timeout.lower() == "none":
-        return
+
     try:
         strategy = CONFIG["WaitStrategy"]
         if strategy.lower() == "legacy":
             logger.debug("Using legacy jQuery-based waiter")
+            if timeout.lower() == "none":
```
turunenm, line 86: Is `timeout` guaranteed to be `str` or can it be type `None`?
Might be a good idea to cast it into a `str` first

QWeb/internal/frame.py
```diff
@@ -78,16 +78,21 @@ def wait_page_loaded() -> None:
...
+                return
             xhr.wait_xhr_legacy(timestr_to_secs(timeout))
         else:
             logger.debug("Using enhanced waiter (network/spinner/render)")
-            xhr.wait_xhr(timestr_to_secs(timeout))
+            if timeout.lower() == "none":
```
turunenm, line 91: Is `timeout` guaranteed to be `str` or can it be type `None`?
Might be a good idea to cast it into a `str` first

QWeb/internal/javascript.py
```diff
@@ -195,513 +111,100 @@ def get_all_elements(css: str) -> list[WebElement]:
...
-        var done = recursiveWalk(node, func);
-        if (done) {
-            return true;
-        }
-        node = node.nextSibling;
-    }
-
-    }"""
+    """Return the recursiveWalk JS function as a string (using external JS file, preloaded)."""
+    return _GET_RECURSIVE_WALK_JS
```
turunenm: If this function is not called outside of this file I would remove it and just refer to the returned "constant" in the functions that call this one.

e.g.
`js = _GET_RECURSIVE_WALK_JS + "\n" + _GET_TEXT_ELEMENTS_FROM_SHADOW_DOM_JS`

At least for me, calling this function raises a suspicion that there's some additional logic / side-effect in it which I need to check to be sure.

QWeb/internal/js/get_all_dropdown_elements_shadow_dom.js
```diff
@@ -0,0 +1,14 @@
+// get_all_dropdown_elements_from_shadow_dom.js
+// Finds all select elements in shadow DOM
+// Assumes recursiveWalk is defined in the context
```
turunenm: Could be useful to tell in the comment where to find the `recursiveWalk`, add to other `recursiveWalk` depending sources if you think it's useful
`// Assumes recursiveWalk (get_recursive_walk.js) is defined in the context`

turunenm (commented):
Looks good with minor remarks

QWeb/internal/frame.py
tkoukkari: that's a good point and based on my understanding it's not prevented anywhere that user could do:
`SetConfig     XHRTimeout    ${None}`

...even if docs specifically say set it to (string) "None" to turn it "off".

Better add some guard / check there. Will fix with a new commit.

QWeb/internal/frame.py
tkoukkari: that's a good point and based on my understanding it's not prevented anywhere that user could do:
`SetConfig     XHRTimeout    ${None}`

...even if docs specifically say set it to (string) "None" to turn it "off".

Better add some guard / check there. Will fix with a new commit.

QWeb/internal/javascript.py
tkoukkari: ok, will change in a new commit.

QWeb/internal/js/get_all_dropdown_elements_shadow_dom.js
tkoukkari: ok

tkoukkari: added this to all js files that need recursive walk

QWeb/internal/javascript.py
tkoukkari: should be changed now.

QWeb/internal/frame.py
tkoukkari: fixed with new commit

QWeb/internal/frame.py
tkoukkari: fixed with new commit

QWeb/internal/search_strategy.py
```diff
@@ -113,8 +113,18 @@ def default_timeout_validation(timeout: Any) -> Any:
         return timeout
 
     @staticmethod
-    def xhr_timeout_validation(timeout: Any) -> Any:
-        return timeout
+    def xhr_timeout_validation(timeout: Any) -> str:
```
turunenm, line 136: Looks good 👍 
Could you check if this is applicable to the `default_timeout_validation` function above and update that also?

turunenm (commented):
lgtm
Would be nice to update `default_timeout_validation` function in the same fashion as you did for the `xhr_timeout_validation`

QWeb/internal/search_strategy.py
tkoukkari: actually it's not good enough, I think we should allow at least this kind of format too: "30s".

I'll change both of those.

tkoukkari: Fixed with a new commit. 

* default_timeout_validation renamed to more generic timeout_validator as it's used in many places
* The new function prevents storing incorrect values
* Robot FW timestr format strings are supported
* Added docstrings to the function to describe what it should do
* xhr_timeout_validation uses the above one, but also handles special case of "none" in all it's capitalization formats

turunenm (approved):

PR #197, opened 2025-11-10
Browser and window helpers

tkoukkari:
- **SwitchBrowser** to support changing by browser instance's **active page title**
- New keyword **ListBrowsers**
- **OpenWindow** to return opened **window handle**
- New keywords **ListWindows** and **GetWindowHandle**
- Modified **SwitchWindow** to work with **window handle** in addition to index. Window handle is preferred in the future
- Improved error message when user tries to open a new browser instance using same profile that is already running
- Updated keyword documentation
- Fixed new issue with Safari 26+, one field needs to be focused before typing


QWeb/internal/browser/__init__.py
```diff
@@ -95,25 +95,34 @@ def get_current_browser() -> WebDriver:
...
-    if str(index).isdigit():
-        if int(index) == 0:
+    if str(target).isdigit():
+        if int(target) == 0:
             raise QWebValueError("SwitchBrowser index starts at 1.")
 
-        i = int(index) - 1
+        i = int(target) - 1
 
         if i < len(_open_browsers):
```
turunenm, line 108: add check for negative out of range index

if there are 3 open browsers and user provides `-4` then `IndexError` raises

turunenm: oh nvm, the initial `if str(target).isdigit():` won't let negative indices through

QWeb/keywords/window.py
```diff
@@ -189,62 +205,126 @@ def close_window() -> None:
 
 @keyword(tags=("Browser", "Interaction", "Window"))
 @decorators.timeout_decorator
-def switch_window(index: str, timeout: Union[int, float, str] = 0) -> None:  # pylint: disable=unused-argument
-    r"""Switch to another tab.
+def switch_window(target: str, timeout: Union[int, float, str] = 0) -> str:  # pylint: disable=unused-argument
+    r"""Switch to another browser tab or window.
+
+    **Preferred usage:**
```
turunenm: Maybe change `Preferred usage` to just `Usage` and then add something like `(preferred usage)` at the end of the first list item.
Current wording / formatting feels bit ambiguous as all items are under the `Preferred usage`

QWeb/keywords/window.py
```diff
@@ -26,6 +27,12 @@
 import platform
 
 
+@dataclass
+class TabInfo:
```
turunenm, line 31: I think it would be better to name this class as `WindowInfo` as these are returned by `List Windows` keyword

Tabs and `TabInfo` makes more sense from the actual user PoV but because existing kws refer to tabs as windows I believe it's better to stick to the existing term for consistency

turunenm (commented):
Code is good, check the comments

QWeb/keywords/window.py
turunenm: `Usage in order of preference` or something like that could lessen ambiguity

samijt (approved):

QWeb/keywords/window.py
tkoukkari: Yes, that was actually my first naming for this dataclass but I think tabs are really what this refers to. However I did not want to rename existing keywords and thought this would highlight that "window" means "tab". 

If this doesn't bother you too much, I would like to keep it as it's visible in docs too (as return type).

QWeb/keywords/window.py
tkoukkari: ah, didn't even think of it that way...the idea there was to be the order of preference, i.e. handle or NEW first, then in simple cases  index...and additionally if index is used make sure what is the index. I'll re-word this as you proposed.

QWeb/keywords/window.py
turunenm: Not a big issue so it's good as is 👍

turunenm (approved):

PR #203, opened 2025-12-10
Rfw 74 support

tkoukkari:
* minor change to keyword censoring logic due to change in internals of rfw 7.4
* Support new "Secret" type in TypeText and TypeSecret keywords

### DIFF
--- a/QWeb/keywords/input_.py
+++ b/QWeb/keywords/input_.py
@@ -32,6 +32,12 @@ from QWeb.internal import element, input_, download, decorators
 from QWeb.internal.input_handler import INPUT_HANDLER as input_handler
 from QWeb.keywords import browser
 
+try:
+    from robot.api.types import Secret
+except ImportError:
+    Secret = str
+
+str_or_secret = Union[str, Secret]
 
 @keyword(tags=("Config", "Input"))
 def set_input_handler(input_method: str) -> None:
@@ -78,7 +84,7 @@ secrets.add_filter("Type Secret3", 1, "hint")
 
 def type_secret3(
     locator: str,
-    input_text: str,
+    input_text: str_or_secret,
     anchor: str = "1",
     timeout: Union[int, float, str] = 0,
     index: int = 1,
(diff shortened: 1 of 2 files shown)


QWeb/keywords/input_.py
```diff
@@ -25,13 +25,21 @@
 from selenium.webdriver.remote.webelement import WebElement
 from robot.api import logger
 from robot.api.deco import keyword
+try:
+    from robot.api.types import Secret
+    BUILT_IN_SECRET = True
+except ImportError:
+    BUILT_IN_SECRET = False
+    Secret = str
```
turunenm: Feel questionable to define the `Secret` type as `str`

I'd prefer defining `str_or_secret` in the try-except and then flip the `isinstance(input_text, Secret)` check

```
try:
    from robot.api.types import Secret
    BUILT_IN_SECRET = True
    str_or_secret = Union[str, Secret]
except ImportError:
    BUILT_IN_SECRET = False
    str_or_secret = str
```

Then on L290 
`if BUILT_IN_SECRET and not isinstance(input_text, str)`

turunenm (changes requested):

QWeb/keywords/input_.py
tkoukkari: Yes, I tried all sorts of ways and I'm pretty sure mypy didn't like this. But I'll check again.

tkoukkari: yup, "error: Cannot assign multiple types to name "str_or_secret""

Kinda of a simple problem but not exactly. I landed on that solution after trying quite a few different ways. From code point of view it's easy but getting mypy to agree, not so much.

tkoukkari: as discussed, I made an alias for "Secret" as this could be potentially valid type somewhere else -> new commit

turunenm: I just realized that the `BUILT_IN_SECRET` bool is unnnecessary if the L290 check is inverted as suggested in the thread start

`if not isinstance(input_text, str)`

tkoukkari: I don't mind it, makes it a bit more easy to see that there is an in-built type as expected, not just that it's not a string

turunenm (approved):

PR #209, opened 2026-04-01
First bidi features

tkoukkari:
Adds:
* Option to open browser in BiDi mode
* Keywords for getting messages, warnings and errors from devconsole
* Acceptance tests for Devconsole keywords
* Keyword docs and doc site example


QWeb/internal/bidi.py
```diff
@@ -0,0 +1,256 @@
...
+
+@dataclass(frozen=True)
+class ConsoleMsg:
+    ts: float
+    level: str
+    text: str
+    url: Optional[str] = None
+    line: Optional[int] = None
+    column: Optional[int] = None
+    source: str = "console"  # 'console' or 'exception'
```
turunenm: better to define this as
`source: Literal['console', 'exception'] = "console"`

QWeb/internal/bidi.py
```diff
@@ -0,0 +1,256 @@
...
+        "critical": "error",  # treat as error
+    }
+    if level in level_map:
+        level = level_map.get(level, level)
+    if level not in valid_levels:
+        raise ValueError(
+            f"Unknown log level: {level}"  # noqa: E501
+        )
+    # Filter by source if specified
+    if source:
```
turunenm: Should check with `is not None`
https://peps.python.org/pep-0008/#programming-recommendations
_Comparisons to singletons like None should always be done with is or is not, never the equality operators._

QWeb/internal/bidi.py
```diff
@@ -0,0 +1,256 @@
...
+    _handler_ids[session_id] = handler_id
+    # Use Selenium's BiDi API for JS exceptions
+    js_handler_id = driver.script.add_javascript_error_handler(
+        lambda event: on_js_exception_event(event, session_id)
+    )
+    _js_handler_ids[session_id] = js_handler_id
+
+
+def _get_console_messages(
+        level: str = "all",
```
turunenm, line 149: Maybe define the levels with `StrEnum` which would eliminate the need for the `level_map`

QWeb/internal/bidi.py
turunenm: reading further this should probably also be defined as `StrEnum` or `Enum` so that it can be used elsewhere to check the value of `source`

QWeb/internal/bidi.py
```diff
@@ -0,0 +1,256 @@
...
+        else:
+            raise ValueError(f"Unknown message source: {source}. Use 'console' or 'js'.")
+    else:
+        filtered_msgs = all_msgs
+    if level == "all":
+        filtered = filtered_msgs
+    else:
+        filtered = [msg for msg in filtered_msgs if _normalize_msg_level(msg, level_map) == level]
+    # Filter by contains if specified
+    if contains:
```
turunenm: `if contains is not None`

turunenm (changes requested):
`Qweb/internal/bidi.py` uses a lot of static string values and the purpose of the same strings across functions are hard to follow.
It's better to use `Enum` or `StrEnum` for such static values to make the intent clearer.

QWeb/internal/bidi.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def on_console_event(event, session_id) -> None:
+    msg = ConsoleMsg(
+        ts=time.time(),
+        level=getattr(event, "level", "all"),
+        text=getattr(event, "message", getattr(event, "text", "")),
+        url=getattr(event, "url", None),
+        line=getattr(event, "line", getattr(event, "lineNumber", None)),
+        column=getattr(event, "column", getattr(event, "columnNumber", None)),
```
turunenm: define `Enum` or `StrEnum` for handled `event` values

QWeb/internal/bidi.py
tkoukkari: fixed by new commit

QWeb/internal/bidi.py
tkoukkari: fixed by new commit

QWeb/internal/bidi.py
tkoukkari: fixed by new commit

QWeb/internal/bidi.py
tkoukkari: fixed by new commit

QWeb/internal/bidi.py
tkoukkari: tbh I'm not sure what all of those are...afaik they are coming from selenium and just forwarded to handler function.

turunenm (approved):

Issue #215, opened 2026-05-15
duties.py doesn't properly escape paths with spaces


billyJoePiano:
### Description of the bug
I am trying to use duty for code quality checks while contributing to the QWeb library, as described here:
https://github.com/qentinelqi/qweb/blob/master/CONTRIBUTING.md

My username on my Windows 11 workstation includes two spaces, and resultingly, the path to my user directory includes two spaces.

Unfortunately, when I run many of the duty commands, I keep getting the following error message: `'C:\Users\<first word in username>' is not recognized as an internal or external command, operable program or batch file`

As I'm looking at the duties.py file in the root of QWeb, I'm noticing there is a failure to properly escape path strings.  I'm working on a fix to this issue, and will include it with my pull request for issue https://github.com/qentinelqi/qweb/issues/214

### To Reproduce
- Install duty on a Windows 11 system, where the user has at least one space in their user name.
- Copy the QWeb repo to a folder within this user's folder: https://github.com/qentinelqi/qweb
- Try to run the requested duty commands, as described in the QWeb contributions document

Example commands (run from cmd.exe) from the directory of the user with at least one space in their name:

`git clone https://github.com/qentinelqi/qweb.git`
`cd .\qweb`
`pip install -r requirements.txt -r requirements_test.txt`

At this point it **may** be necessary to update the Window PATH environmental variable as described in the pip command output.  Go to Start -> Edit the system environment variables -> Environment Variables (button) -> Path (selection) -> Edit (button), and proceed update the PATH variable as instructed by pip output.  Then restart the command prompt, and navigate back to the qweb repo directory

`pre-commit install`
`duty format`
`duty lint`
`duty typing`
`duty unit-tests` (this is the only duty command which actually worked for me)
`duty acceptance-tests`

### Expected behavior
`ctx.run(...)` calls (within duties.py) should properly escape spaces for shell commands, by using a list of arguments rather than a single string with spaces separating the arguments.

### Environment information

```bash
duty --debug-info  # | xclip -selection clipboard
```

- __System__: Windows-11-10.0.26200-SP0
- __Python__: cpython 3.12.10 (C:\Users\my user name\AppData\Local\Python\pythoncore-3.12-64\python.exe)
- __Environment variables__:
- __Installed packages__:
  - 'duty' v1.9.0

Issue #214, opened 2026-05-13
XHR Monitor bugs


billyJoePiano:
**Describe the bug**
The script install_monitor.js sometimes gets stuck, thinking there are unresolved pending XHR requests which in actuality have been garbage-collected or aborted.  This can lead to unnecessary delays in QWeb keywords which wait (upto a timeout) for the `window.__xhrMon.pending` count to reach 0 before proceeding.

I've dug into the XHR problem, and identified two root causes which are closely related to each other:
1) XHR requests which are sent (and reach readyState = 1 immediately after sending), but are garbage-collected before completing.  Presumably, these requests were abandoned by the code which created them, leaving no strong references remaining in the global scope.
2) XHR requests which are sent (and reached readyState = 1 immediately after sending), but inexplicably revert to readyState = 0 without ever invoking the "loadend" event listener.

In both cases, the "loadend" event listener is never invoked, and consequently the `window.__xhrMon.pending` number is never decremented, leaving the XHR Monitor hanging, believing there are still unresolved pending XHR request(s).

Based on my experiments, it seems an XHR request which may be subject to one of the above scenarios can also be prone to the other.  It is simply a question of whether the JS engine's garbage-collector gets to it first, or the readyState reverts to 0 first.

**To Reproduce**
1. Enable debugging in the install_monitor.js script, (on line 6, set `debug = true`)
2. Start a robotic script in debug mode, which opens the web browser, navigates to www.google.com, and calls some QWeb keyword to install the XHR Monitor (e.g. `Verify Text      About`)
3. Open the browser dev console to observe the debug logs
4. Return to the browser and start typing in the search box
5. With each keystroke a new XHR is sent out.  However, if you press multiple keys in rapid succession, a new XHR is sent before the previous ones returns.  There is something about Google's client-server interaction which causes the old XHR to be dropped without ever triggering the "loadend" event listener.  This causes the `window.__xhrMon.pending` to never return to 0.

**Expected behavior**
The `window.__xhrMon.pending` value should accurately reflect the number of outstanding XHR requests still pending.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome 148.0.7778.97
 - Chromedriver version: 148.0.778.167
 - QWeb Version 3.7.4

**Additional context**

I have already written a patch to this problem, which I will submit as a pull request soon.  Additionally, there a few other minor issues I identified in the XHR Monitor, which I will fix in my PR.

[...]

PR #216, opened 2026-05-15
Fixes to XHR monitor, duties.py

billyJoePiano:
This includes fixes to issues described in:

- https://github.com/qentinelqi/qweb/issues/214
- https://github.com/qentinelqi/qweb/issues/215

All unit tests and acceptance tests passed, running on my local system (Windows 11 with Python 3.12.10)

### DIFF
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -3,7 +3,7 @@
 // Always returns true on success (even if some patches are skipped).
 // The optional `debug` flag (default: false) is used only for troubleshooting.
 // ...full code below...
-return (function (debug = true) {
+return (function (debug = false) {
 	if (window.__xhrMon && window.__xhrMon.installed) return true;
 
 	// to circumvent any website monkey-patching of native functions
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -3,81 +3,435 @@
 // Always returns true on success (even if some patches are skipped).
 // The optional `debug` flag (default: false) is used only for troubleshooting.
 // ...full code below...
-return (function (debug = false) {
+return (function (debug = true) {
 	if (window.__xhrMon && window.__xhrMon.installed) return true;
 
-	window.__xhrMon = Object.assign(window.__xhrMon || {}, {
+	// to circumvent any website monkey-patching of native functions
+	const _console = {
+		log: console.log,
+		warn: console.warn ? console.warn : console.log,
+		error: console.error ? console.error : console.log
+	};
+	const Array_from = Array.from;
+	const setTimeout = window.setTimeout;
+
+	//use polyfills if neccessary
+	const Object_getPrototypeOf = Object.getPrototypeOf ? Object.getPrototypeOf : function(obj) { return obj.__proto__; };
+	const performance_now = window.performance && typeof window.performance.now === "function" ? window.performance.now.bind(window.performance) : window.Date.now.bind(window.Date);
+	const Set = window.Set ? window.Set : SetPolyfill();
+	const FinalizationRegistry = window.FinalizationRegistry ? window.FinalizationRegistry : FinalizationRegistryPolyfill();
+	const WeakRef = window.FinalizationRegistry && window.WeakRef ? window.WeakRef : WeakRefPolyfill();
+	const WeakSet = window.WeakSet ? window.WeakSet : WeakSetPolyfill();
+	const WeakMap = window.WeakMap ? window.WeakMap : WeakMapPolyfill();
+	const Symbol = window.Symbol ? window.Symbol : Object;
+
+	const mon = window.__xhrMon = window.__xhrMon ? window.xhrMon : {
 		installed: true,
-		pending: (window.__xhrMon && typeof window.__xhrMon.pending === "number") ? window.__xhrMon.pending : 0,
+		pending: window.__xhrMon && typeof window.__xhrMon.pending === "number" ? window.__xhrMon.pending : 0,
 		fetchPatched: false,
 		xhrPatched: false,
-		lastMutationTs: Date.now(),
-		observerStarted: false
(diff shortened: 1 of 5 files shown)


tkoukkari:
Can you please separate duty issue and xhr monitor issue to separate PR's? Easier to handle.

My first thought is that the XHR one changes too much and adds too much complexity for a corner case we've never really seen in real world usage. 

Original script is simple: increment pending on send, decrement on loadend.

New version adds WeakRef, FinalizationRegistry, custom Set/WeakMap polyfills, monkey patch protection, periodic polling, and computed pending state. That is a large risk surface for a monitor script. 

I'm not sure if Finalization callbacks are a good main mechanism for unblocking QWeb waits. They are not guaranteed to run promptly, or even at all. 

I do agree that using permance.now would be better here.

Do you have an reproducible example test case that could show the problem (or that it actually blocks usage)? I could not make this break with the google example given above.

billyJoePiano:
> Can you please separate duty issue and xhr monitor issue to separate PR's? Easier to handle.

Sure.

> My first thought is that the XHR one changes too much and adds too much complexity for a corner case we've never really seen in real world usage.
> 
> Original script is simple: increment pending on send, decrement on loadend.
> 
> New version adds WeakRef, FinalizationRegistry, custom Set/WeakMap polyfills, monkey patch protection, periodic polling, and computed pending state. That is a large risk surface for a monitor script.

I added the Polyfills because I wasn't sure how much backwards compatibility with older browsers we want to maintain.  If we can assume QWeb will only be used on a modern (after ~2020) standards-compliant browser, that could help simplify the JS somewhat.  I'm just not certain whether QWeb is used to test legacy browsers/browser-versions, and/or if some of the mobile browsers don't support all of the same JS features as a desktop browser.

> I'm not sure if Finalization callbacks are a good main mechanism for unblocking QWeb waits. They are not guaranteed to run promptly, or even at all.

The Finalization callback is not really the main mechanism for unblocking waits.  Finalization is there as a fallback, because the "loadend" event listener wasn't always triggering prior to garbage collection of the XHR.

With my new JS, there are three main mechanisms for catching when an XHR is no longer pending, in order of precedence:

1. Event listener on XHR itself.  In the old code this was added to the "loadend" event.  After some experimentation, I changed it to "readystatechange" event which seemed to improve reliability (explanation below)
2. Polling the list of pending XHR Weak References, to confirm a) they haven't been garbage collected (i.e. the WeakRef dereferences to undefined), and b) haven't reverted to readyState = 0 while bypassing the event listener
3. FinalizationRegistry.  In a modern browser with native WeakRef, this is technically redundant with 2.a) above, since that would also indicate garbage collection.  However, there is more of a delay for approach 2.a), which is why I left the FinalizationRegistry in place, because it tends to trigger immediately upon garbage collection whereas 2.a) doesn't find out until the next polling.

In practice, when I was developing my new JS, the finalization registry was triggered the least of the 3 methods above.   The event listener on the XHR itself is by far the most common, but scenario 2.b) happened fairly frequently, with 3. happening the least but still occasionally.  I never observed scenario 2.a) because the FinalizationRegistry would always get triggered first, before the polling would catch a GC'd WeakRef.

Interestingly, once I changed the event listener from the "loadend" event to the "readystatechange" event, I never encountered scenario 3 at all, and I'm not sure whether I saw scenario 2.b) either.  It's possible that was the real root cause of the issue, but I think we'd still want to keep check 2. as a fallback in case the listeners fail for whatever reason.

> I do agree that using permance.now would be better here.

Related to the earlier concern about Polyfills and needless complexity... note that I did include Date.now as a fallback in case performance.now doesn't exist.  If that is making the code too complicated, we can make the assumption that performance.now is always available.  Thoughts?

> Do you have an reproducible example test case that could show the problem (or that it actually blocks usage)? I could not make this break with the google example given above.

The google.com search box is the only way I've been able to reproduce it reliably.  (Note: I'm using the most recent version of Google Chrome on a Windows 11 system)

However, in my QA automation work (mostly in the Salesforce eco-system, using Copado Robotic Testing) I have definitely experienced sporadic delays where QWeb keywords take far longer than expected to complete.  And once encountered, these delays seem to happen with every subsequent QWeb invocation until a new page is loaded.

I can't prove those delays were because of the XHR monitor bug, because a) the debug flag is typically disabled, and b) RF test scripts typically don't capture JS console logs.  But those delays are consistent with what we'd expect based on my trials with the old XHR monitor code.  So I have a strong inkling this bug is happening and causing delays more often than you realize.  It is just sporadic and difficult to reproduce.

My goal with that JS was to make it absolutely iron-clad, but I realize that comes at the cost of some complexity which can impact readability and maintainability.  If you'd like, I can start over with original JS, and try to make a more minimalist version of this script that doesn't include Polyfills or FinalizationRegistry, and try to keep the approach simpler where possible, at the expense of possible backwards compatibility issues and problems stemming from unexpected browser and/or website monkey-patching behavior.

That said, I think it will still be important to
1. Poll WeakRefs to the XHRs, to confirm against garbage-collection and readyState reversion
2. Have a computed pending count, to ensure each tick in the count represents a real XHR that actually exists and is in a pending readyState
3. Protect against monkey-patching of the addEventListener method, since that is that is very common method to monkey-patch by various front-end frameworks and SaaS websites

Thoughts?

billyJoePiano:
@tkoukkari 

One other thing I'll mention, for what it's worth -- I did test out the polyfills extensively using Chrome, to make sure they didn't cause any issues with the code.  Everything worked exactly how it was supposed to, and should (in theory) enable the JS to work on pre-2015 browsers (perhaps except for IE)

That is to say -- I'm very confident that my JavaScript is air-tight.  But I also recognize that it is more confusing and complicated, which could be difficult to maintain long-term.  I'm submitting another PR shortly, with a simplified version of that code, which doesn't include all of the features while still solving the core problem.

tkoukkari:
Closing this one as this PR was split into two. Let's review the most recent ones.

PR #218, opened 2026-05-15
XHR monitor fixes, v2 (simplified)

billyJoePiano:
This is a follow-up to feedback in the comments of https://github.com/qentinelqi/qweb/pull/216

It fixes the issues described in https://github.com/qentinelqi/qweb/issues/214

All unit-tests and acceptance-tests passed on my local system (Windows 11, Python 3.12.10)

I have re-worked my JavaScript (from PR 216) to simplify it, for the sake of clarity and future maintainability, while still solving the core problem described in Issue 214.

In contrast to my other PR, this JavaScript:
- Does not require the use of Set, WeakSet, WeakMap, or FinalizationRegistry.  The only "modern" browser feature required is WeakRef
- Does not include extensive protections against websites that monkey-patch native methods.  As a result, there is a small chance of unexpected behavior if a website overrides or wraps certain native APIs (e.g. array.push, array.splice, function.apply, console.log, setTimeout, promise.finally, etc...).  *
- Does NOT include Polyfills and legacy syntax (e.g. this script uses 'let' instead of 'var' ... which was already the case with the script it replaces in the main branch).  As a result, it may encounter errors on legacy browser versions, or mobile browsers which don't have the full modern desktop browser APIs.

*Note that I did include monkey-patch protection for two functions critical to the purpose of the script -- xhr.addEventListener() and xhr.readyState getter.  The former is exceedingly common for front-end frameworks and SaaS websites to monkey-patch, so these seemed critical to protect.

Please let me know your thoughts.  Thank you!

### DIFF
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -42,7 +42,7 @@ return (function (debug = false) {
 					/* In case the finally/then callback function runs synchronously upon the call to promise.finally()/promise.then() ...
 					 * The symbol is constructed first, so it is available to the closure(s), but fetchDone is scheduled async via setTimeout.
 					 * This is because fetchDone should not run until AFTER fetchPending array has been populated with the symbol (below), yet we don't
-					 * want to populate it with the symbol until promise.finally/promise.then has completed successfully. */
+					 * want to populate it with the symbol until promise.finally/promise.then has completed successfully (i.e. without throwing an exception...) */
 					const symbol = Symbol();
 					if (typeof promise.finally === "function") {
 						promise = promise.finally(function() { setTimeout(fetchDone, 0, symbol); });
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -115,7 +115,6 @@ return (function (debug = false) {
 						console.error("XHR monitor: Error in xhr.open:", e);
 					}
 				}
-
 				return returnVal;
 			}
 
@@ -200,7 +199,10 @@ return (function (debug = false) {
 							if (debug) {
 								console.warn("XHR monitor: already-sent request was garbage collected without triggering listener, pending count:", window.__xhrMon.pending);
 							}
-						} else if (_getReadyState.apply(xhr) === 0) {
+							continue;
+						}
+						const readyState = _getReadyState.apply(xhr);
+						if (readyState === 0 || readyState === 4) {
 							xhr.__xhrDone = true;
 							xhrPending.splice(i, 1);
 							if (debug) {
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -3,7 +3,7 @@
 // Always returns true on success (even if some patches are skipped).
 // The optional `debug` flag (default: false) is used only for troubleshooting.
 // ...full code below...
-return (function (debug = false) {
+return (function (debug = true) {
 	if (window.__xhrMon && window.__xhrMon.installed) return true;
 
 	window.__xhrMon = Object.assign(window.__xhrMon || {}, {
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -3,7 +3,7 @@
 // Always returns true on success (even if some patches are skipped).
 // The optional `debug` flag (default: false) is used only for troubleshooting.
 // ...full code below...
-return (function (debug = true) {
+return (function (debug = false) {
 	if (window.__xhrMon && window.__xhrMon.installed) return true;
 
 	window.__xhrMon = Object.assign(window.__xhrMon || {}, {
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -3,7 +3,7 @@
 // Always returns true on success (even if some patches are skipped).
 // The optional `debug` flag (default: false) is used only for troubleshooting.
 // ...full code below...
-return (function (debug = false) {
+return (function (debug = true) {
 	if (window.__xhrMon && window.__xhrMon.installed) return true;
 
 	window.__xhrMon = Object.assign(window.__xhrMon || {}, {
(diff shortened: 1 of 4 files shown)


tkoukkari:
Thanks! We will go through this and I'll comment here once we have. Note that it may take a week or two as this week is pretty busy.

tkoukkari:
TBH I'm not sure about this. The originating issue is a corner-case where loadend may not always fire, which afaik is not normal XHR behavior and something we have not been able to reproduce ourselves.

 What I'm not so sure about is:
* The change is making quite broad changes to the whole underlying architecture, which are not all necessary based on the original issue
* My main concern is the CDP-based monitor injection in browser.py via xhr.py. This is much riskier than the XHR-monitor cleanup itself, because it changes browser startup behavior for Chromium sessions and assumes Page.addScriptToEvaluateOnNewDocument is available. QWeb already supports remote Chrome/Edge sessions through generic RemoteWebDriver paths, including Grid and BrowserStack, and those paths may not reliably support this CDP call. In that case, this change could break OpenBrowser for existing supported configurations, which is a much more concrete regression risk than the original issue.
* Also non-Chromium based browsers behaving differently complicates testing etc.

All in all, I feel this fixes a corner case issue most won't ever see with quite drastic changes that could potentially cause regression and at least would warrant huge amount of testing in different scenarios.

The Date.now() -> performance.now() change makes perfect sense and should be done.

But I'll think about this a bit more and let other reviewers voice their opinion, especially if it's different from mine.

billyJoePiano:
> TBH I'm not sure about this. The originating issue is a corner-case where loadend may not always fire, which afaik is not normal XHR behavior and something we have not been able to reproduce ourselves.

Hi @tkoukkari .  I will record a video demonstrating how I reproduce this defect.  Hopefully that will help convince you of the necessity of this fix.

I use QWeb extensively in my job (via Copado Robotic Testing), and have noticed sporadic delays in the execution of QWeb keywords, where the keyword takes far longer than normal to complete (e.g. about 30 seconds, when it should take less than 1 second).  Once encountered, these 30 second delays seem to happen with every subsequent QWeb invocation until a new page is loaded.  This behavior is consistent with the defect in the XHR monitor I've identified and fixed here.


> What I'm not so sure about is:
> 
> * The change is making quite broad changes to the whole underlying architecture, which are not all necessary based on the original issue

Which changes do you think are not necessary based on the original issue?


> * My main concern is the CDP-based monitor injection in browser.py via xhr.py. This is much riskier than the XHR-monitor cleanup itself, because it changes browser startup behavior for Chromium sessions and assumes Page.addScriptToEvaluateOnNewDocument is available. QWeb already supports remote Chrome/Edge sessions through generic RemoteWebDriver paths, including Grid and BrowserStack, and those paths may not reliably support this CDP call. In that case, this change could break OpenBrowser for existing supported configurations, which is a much more concrete regression risk than the original issue.

That is an understandable concern.  I can wrap the CDP call in try/except, so that it doesn't cause any problems with OpenBrowser.


> * Also non-Chromium based browsers behaving differently complicates testing etc.

The only difference is when the xhr monitor script loads -- on new document (chromium) vs. on the first QWeb keyword invocation (non-chromium).  The script itself is idempotent, so if there are multiple attempts to load it, the second and subsequent calls have no effect.

If you think this difference in timing of script load will significantly complicate regression testing of QWeb, I'm amenable to removing the CDP call altogether.


> All in all, I feel this fixes a corner case issue most won't ever see with quite drastic changes that could potentially cause regression and at least would warrant huge amount of testing in different scenarios.

Based on my experience using QWeb, I'm quite confident we are seeing this issue on a regular basis.  While it doesn't trigger failures or errors (...probably the reason it wasn't reported sooner...) it does cause unnecessary delays in test script execution.

billyJoePiano:
@tkoukkari 

I have uploaded a demonstration video of the bug.  Apologies that the audio is very quiet, you may need to turn up the volume a bit to hear me talking.  See the below link:

https://www.youtube.com/watch?v=_bCTeaOuPgs

Also, as you can see from the above commit, I removed the CDP callout since that seemed to be your biggest concern.  Please let me know your thoughts, and if you have any other concerns.  Thanks.

tkoukkari:
Thanks again for digging into this and for the detailed investigation. It was genuinely helpful.

I was eventually able to reproduce the issue on my side as well. I may be wrong, but at least from what I saw it seems related to cancelled, aborted, or timed out XHR requests leaving the monitor in a pending state.

Because of that, I tried a much narrower alternative fix that only touches the JS monitor code, to see whether the issue could be addressed without changing the broader browser startup / monitor-install flow. With that narrower version I wasn’t able to reproduce the problem anymore.

Would you mind taking a look and checking whether this also covers the cases where you originally saw the issue?
[install_monitor.js](https://github.com/user-attachments/files/28254134/install_monitor.js)
[wait_status_lite.js](https://github.com/user-attachments/files/28254135/wait_status_lite.js)

The main reason I’m suggesting this is just to see whether a smaller fix would be enough here, or whether we would still be missing an important case from your original PR.

If this narrower version looks good to you as well, I’d be very happy for you to update your PR in that direction. You were the one who found the issue and put together the original fix proposal, so I’d definitely prefer to keep you as the contributor on this one.

And in any case, thanks again for the report and the work here. Contributions like this are very welcome.

billyJoePiano:
Thanks for your response @tkoukkari 

I'm looking at your code, and I see a few areas where it could potentially "leak" pending XHR ticks that never get decremented.  I haven't tested it yet, so these are just hypothetical scenarios.  It might take extensive testing to determine whether these "leaks" happen in real-world scenarios.  However, it would definitely be possible to reproduce some of these scenarios with a custom website.

1. **Multiple calls to `xhr.send()`.**  If `xhr.send()` is called multiple times, it will result in multiple increments/ticks to the pending count, but only one decrement after it resolves.  Presumably, the native method called on line 80 will throw an exception on the second and subsequent times it gets called, but that exception happens _after_ the increment of the pending count.  This concern is the main reason I put the native call to `xhr.send` prior to the monitor's incrementing logic in my version of the code, so that any exception is thrown prior to incrementing the count.  Additionally, I included a tracking field `this.__xhrSent` which is populated when the xhr is sent, so we are not double-tracking any xhrs.

2. **Monkey patching of `addEventListener()`.**  It is exceedingly common to see monkey-patching of the method `this.addEventListener()` by various front-end frameworks and SaaS websites.  I have personally witnessed cases where the monkey-patched `addEventListener()` method acts as a filter, and refuses to add event listeners it doesn't recognize (e.g. from user scripts, like ours).  I couldn't tell you the reasons why these websites do this, but I feel it would be a mistake not to account for this potential behavior in our code.

3. **If native fetch and/or promise API throws an exception.**  Analogous to the order-of-operations issue explained in 1. above, the pending count is incremented (line 22) _before_ confirming that the native fetch method doesn't throw an exception (line 23).  If an exception is thrown at line 23 when invoking the native fetch (or subsequently in lines 24-27 when installing the promise call-backs), then that tick in the pending count will never get decremented.

There's also some minor potential side-effects of this code (which is also true of my v2), where properties are being added to the XHR object itself (e.g. `this.__xhrTracked`) that are enumerable and visible to the websites' own JS, and therefore could potentially result in unexpected behavior for the website.  Additionally, in your version of the code, line 57 adds a persistent strong reference to the xhr that could prevent its garbage collection (thus causing a memory leak), since that variable is referenced in the `done` closure that potentially remains referenced from the global scope (via the event listeners) even after the xhr has finished.

I realize the risk from these side-effects is minor, but it was part of the reason why my v1 of the code used WeakSet and WeakMap for tracking (as opposed to fields on the object itself, e.g. `this.__xhrTracked`) and WeakRef (as opposed to strong references for the closures), to minimize any undesirable side-effects, especially those visible to the website's JS itself.

Finally, the `readystatechange` event listener is checking for `xhr.readyState === 4 || xhr.readyState === 0`, however when you add this event listener it never actually checks that ready state has advanced past 0.  Again, a relatively minor issue, but something I wanted to button-up to prevent any unforseen/undefined behavior in the future.

All this said, I realize your goal here is to fix the problem with as minimal changes as needed, for the sake of readability and maintainability.  I can make a v3 of my code that takes into account your approach.  I will let you know when that PR is ready.

billyJoePiano:
Version 3 Pull Request is ready:

https://github.com/qentinelqi/qweb/pull/220

tkoukkari:
Closing this one, replaced by #220

PR #220, opened 2026-05-26
XHR Monitor fixes, v3 (bare-bones)

### LINKED ISSUE #214: XHR Monitor bugs

billyJoePiano:
This is version 3 of my proposed fix to the XHR Monitor, based on previous discussions in Pull Requests:
https://github.com/qentinelqi/qweb/pull/218
https://github.com/qentinelqi/qweb/pull/216

It solves the bug described in issue https://github.com/qentinelqi/qweb/issues/214

All unit-tests and acceptance-tests passed on my local system (Windows 11, Python 3.12.10)

This v3 proposal is the most bare-bones possible fix I'd recommend using.  Anything less and there would be considerable edge-case scenarios where the XHR Monitor could break (see also: https://github.com/qentinelqi/qweb/pull/218#issuecomment-4545678323 )

### DIFF
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -45,12 +45,17 @@ return (function (debug = false) {
 				const returnVal = _open.apply(this, arguments);
 				try {
 					if (!this.__xhrOpened) {
-						_addEventListener.call(this, "loadend", done, { once:true });
-						_addEventListener.call(this, "abort", done, { once:true });
-						_addEventListener.call(this, "error", done, { once:true });
-						_addEventListener.call(this, "timeout", done, { once:true });
+						_addEventListener.call(this, "loadend", done);
+						_addEventListener.call(this, "abort", done);
+						_addEventListener.call(this, "error", done);
+						_addEventListener.call(this, "timeout", done);
 						_addEventListener.call(this, "readystatechange", readystatechange);
 						this.__xhrOpened = true;
+					} else {
+						// In case of re-opened XHR, consider the previous request done
+						done.apply(this); //automatically checks if this was actually pending before decrementing
+						this.__xhrSent = false;
+						this.__xhrDone = false;
 					}
 				} catch(e) {
 					if (debug) console.error("XHR monitor: error in open", e);
@@ -103,6 +108,9 @@ return (function (debug = false) {
 			function readystatechange() {
 				const readyState = _getReadyState.apply(this);
 				if (readyState === 4 || readyState === 0) {
+					// readyState === 4 means DONE
+					// readyState === 0 means UNSENT and can also mean aborted or failed
+					// In both cases we consider the request done
 					done.apply(this);
 				}
 			}
--- a/QWeb/internal/js/install_monitor.js
+++ b/QWeb/internal/js/install_monitor.js
@@ -11,67 +11,109 @@ return (function (debug = false) {
 		pending: (window.__xhrMon && typeof window.__xhrMon.pending === "number") ? window.__xhrMon.pending : 0,
 		fetchPatched: false,
 		xhrPatched: false,
-		lastMutationTs: Date.now(),
+		lastMutationTs: performance.now(),
 		observerStarted: false
 	});
 
 	try {
 		if (!window.__xhrMon.fetchPatched && typeof window.fetch === "function") {
-			const _orig = window.fetch;
+			const _fetch = window.fetch;
 			window.fetch = function() {
-				try { window.__xhrMon.pending++; } catch(e) {}
-				const p = _orig.apply(this, arguments);
-				const dec = function(){ try { window.__xhrMon.pending--; } catch(e) {} };
-				return p && typeof p.finally === "function"
-					? p.finally(dec)
-					: p.then(function(r){ dec(); return r; }, function(e){ dec(); throw e; });
+				let p = _fetch.apply(this, arguments);
+				try {
+					const dec = function(){ try { window.__xhrMon.pending--; } catch(e) {} };
+					p = p && typeof p.finally === "function"
+						? p.finally(dec)
+						: p.then(function(r){ dec(); return r; }, function(e){ dec(); throw e; });
+					window.__xhrMon.pending++;
+				} catch(e) {
+					if (debug) console.error("XHR monitor: error in fetch", e);
+				}
+				return p;
 			};
 			window.__xhrMon.fetchPatched = true;
 		}
 	} catch(e){if (debug) console.warn("XHR monitor: fetch patch failed", e);}
 
-		try {
+	try {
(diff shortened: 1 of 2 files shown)


tkoukkari:
@billyJoePiano Thanks, this is much closer to the kind of narrow-scope fix I had in mind.

I like that this keeps the change limited to the JS monitor logic and avoids the broader browser-startup / CDP changes from the earlier version. On a quick review this looks much more in line with what I’d be comfortable merging.

I still want to do a bit more testing on my side before I approve it, mainly just to make sure it behaves well in practice and that we are not introducing any new wait-related edge cases.One small thing I noticed: if the same XMLHttpRequest object is reused across multiple open() / send() cycles, I think the current __xhrSent / __xhrDone flags may carry over from the first request. In that case, the later request might not get tracked correctly. It seems like the simplest fix would be to reset those per-request flags in open(), while still attaching the listeners only once.

Overall though, this looks much closer to the scope I was hoping for. I’m planning to close the previous two PRs related to this issue, and treat this one as the replacement PR that we continue evaluating.

QWeb/internal/js/install_monitor.js
```diff
@@ -11,67 +11,109 @@ return (function (debug = false) {
...
+					if (debug) {
+						console.log("XHR monitor: request ended, pending count:", window.__xhrMon.pending);
+					}
+				} catch(e) {
+					if (debug) console.error("XHR monitor: error in done", e);
+				}
+			}
+			function readystatechange() {
+				const readyState = _getReadyState.apply(this);
+				if (readyState === 4 || readyState === 0) {
```
turunenm, line 110: Add a comment what these `readyState`s are and/or add link to documentation where the states are described.
Ideally these should be checked against an enum but given the nature of the monitor a comment will work just fine.

billyJoePiano:
> One small thing I noticed: if the same XMLHttpRequest object is reused across multiple open() / send() cycles, I think the current __xhrSent / __xhrDone flags may carry over from the first request. In that case, the later request might not get tracked correctly. It seems like the simplest fix would be to reset those per-request flags in open(), while still attaching the listeners only once.

Interesting.  I didn't realize an XHR could be reused, but I looked it up and you are correct.  It's a non-standard usage, but it can be done.  I'll adjust the code accordingly.

> Overall though, this looks much closer to the scope I was hoping for. I’m planning to close the previous two PRs related to this issue, and treat this one as the replacement PR that we continue evaluating.

Thanks

billyJoePiano:
Hi @tkoukkari -- I just added a few lines in the JS to deal with the XHR re-use scenario.  I noticed that the checks have already started, so you may want to re-run the checks with the new code.

tkoukkari (approved):
Thanks for iterating on this and for working through the edge cases. This latest version looks good to me.

As far as I can tell, and to the best of my ability to test this so far, everything now looks good and performs as expected. I like that the fix stays narrowly scoped to the JS monitor logic and avoids the broader browser-startup / CDP changes from the earlier versions. The reused-XHRs concern also looks addressed now.

I ran additional testing on my side, including the cases we discussed, and it behaved correctly in those tests as well as the other checks I ran.

Approved from my side.

Thanks again for digging into this and for sticking with the iterations here. This was a very helpful contribution!

billyJoePiano:
> Thanks again for digging into this and for sticking with the iterations here. This was a very helpful contribution!

Thank you!  Glad to contribute!  I hope this contribution proves useful.

turunenm (approved):

PR #228, opened 2026-08-14
fix: addes support for browser options/flags that contain ","

tkoukkari:


test/unit/test_utils.py
```diff
@@ -67,3 +67,43 @@ def test_xpath_validator():
...
+
+def test_option_handler_splits_options():
+    result = option_handler("--kiosk, --disable-gpu")
+    assert result == ["--kiosk", "--disable-gpu"]
+
+
+def test_option_handler_preserves_quoted_commas():
+    result = option_handler(
+        '"--host-resolver-rules=MAP * ~NOTFOUND , EXCLUDE client-proxy.copatunnel.svc.cluster.local", '
+        '--proxy-server=socks5://client-proxy.copatunnel.svc.cluster.local:54321'
```
turunenm: I'd remove the `copatunnel.svc.cluster` part from the test uris

turunenm (changes requested):

test/unit/test_utils.py
tkoukkari: ah, it was still left there. I removed it from the kw doc examples but forgot them being there as well. Will update.

tkoukkari: Fixed with new commit

turunenm (approved):

PR #231, opened 2026-09-11
feat: adds env variable "TEST_WORKSPACE_ROOT" as one of the search co…

tkoukkari:
…nditions of "default folders".

+ Overall better documentation how default folders are searched.


QWeb/internal/download.py
```diff
@@ -126,7 +154,21 @@ def get_path(filename: str) -> Path:
     exec_dir = BuiltIn().get_variable_value("${EXECDIR}")
     files_exec_dir = Path(f"{get_exec_subdir(exec_dir, 'files')}/{filename}")
     images_exec_dir = Path(f"{get_exec_subdir(exec_dir, 'images')}/{filename}")
-    paths = [downloads, files, images, files_exec_dir, images_exec_dir]
+    test_workspace_root = os.environ.get("TEST_WORKSPACE_ROOT")
+    if test_workspace_root:
```
turunenm, line 163: I assume that you want have a specific order for the paths, so I would do it like this:

```
paths = [downloads, files, images]
if test_workspace_root:
  paths.extend([test_workspace_files, test_workspace_images])
paths.extend([files_exec_dir, images_exec_dir])
```

I think it's easier to grasp that the ordering is meaningful and also which paths land in the list

QWeb/keywords/icon.py
```diff
@@ -44,18 +44,14 @@ def click_icon(
...
-
     Examples
     --------
     .. code-block:: robotframework
 
         *** Variables ***
         ${BASE_IMAGE_PATH}          ${CURDIR}${/}..${/}resources${/}images
 
-    BASE_IMAGE_PATH should lead to the folder where all your reference icons are
+    ``${base_image_path}`` should lead to the folder where reference icons are stored.
```
turunenm: Var name `${BASE_IMAGE_PATH}` in UPPER case to match the var declaration in the example, apply to all instances

docs/how_to/upload_download.md
```diff
@@ -37,37 +37,109 @@ UploadFile    Upload Document    invoice.pdf
...
-- `${EXECDIR}/**/files`
+#### Search order
 
-This allows you to avoid hardcoded absolute paths.
+If the given name is not already an existing path, QWeb searches in this order:
 
-### Why Default Folders Matter
+1. User Downloads folder
+2. Suite-local `files/` and `images/` near `${SUITE SOURCE}` (see below)
+3. `files/` and `images/` anywhere under `TEST_WORKSPACE_ROOT` (when set)
```
turunenm: use `%{TEST_WORKSPACE_ROOT}` which is the RF syntax for env vars, apply to all user facing docstring instances

turunenm (changes requested):

docs/how_to/upload_download.md
tkoukkari: should be fixed with a new commit

QWeb/keywords/icon.py
tkoukkari: should be fixed with a new commit

QWeb/internal/download.py
tkoukkari: should be fixed with a new commit

turunenm (approved):

